PR #9, opened 2017-05-11
fix(proguard): adding -dontshrink param to proguard to avoid removing…

milos1290:
… unused classes


benmarten:
in unused, I assume you mean as unused detected by pro guard or android studio?

PR #10, opened 2017-05-12
Remove fire base related code

cnevinc:
Since we are not using Firebase, I want to remove firebase related code in the code base.


benmarten:
I don't think we can merge this pr like this, as this would remove firebase completely. We need to move firebase and gcm to a separate module completely.

PR #11, opened 2017-05-15
fix(NPE):  fix NPE at VarCache.java.

Enuviel:


AndroidSDK/src/com/leanplum/internal/VarCache.java
```diff
@@ -465,14 +465,16 @@ static int getResIdFromPath(String resPath) {
...
       Var<?> var = vars.get(name);
-      String overrideFile = var.stringValue;
-      if (var.isResource && (var.kind().equals(Constants.Kinds.FILE)) && overrideFile != null &&
-          !var.defaultValue().equals(overrideFile)) {
-        Map<String, Object> variationAttributes = CollectionUtil.uncheckedCast(fileAttributes.get
-            (overrideFile));
-        InputStream stream = fileStreams.get(overrideFile);
-        if (variationAttributes != null && stream != null) {
-          var.setOverrideResId(getResIdFromPath(var.stringValue()));
+      if (var != null) {
```
benmarten: a more elegant solution could be `if (var == null) { continue; }`

milos1290:
Shouldn't be continue instead of retrun ?

Enuviel:
yeah, thanks!

PR #13, opened 2017-05-17
Strip unused files or depedencies

cnevinc:
This PR is just for your reference. 
Since we can't support build from Gradle directly, we can't just use this SDK as a module and build it with Gradle. 
Key changes are
1. Remove FCM related code and entries
2. Remove Geolocation related code and entries
3. Change the HttpClient to our fork version (didn't add this to your Gradle file)
4. Source code will be copy and paste to our repository. The changes in build.gradle and AndroidManifest.xml is just a proposal. Please find https://bugzilla.mozilla.org/show_bug.cgi?id=1351585 for details.
5. We already have a GcmListenerService so may need to do some fix about it.


benmarten:
Thanks for providing PR. Unfortunately we will not be able to merge this PR, as this will remove FCM, rather than refactoring to splitting it into a separate module. Also critical configuration from build.gradle is removed.

I am still not happy about your approach to using open source software. Of course you are free to do whatever you want with the source code, however I see several drawbacks with your approach:
- Since you are using a modified SDK it has not been qa'ed like our official releases.
- We explicitly state that we cannot support custom forks at this point.
- Any future updates to the SDK are hard to get into your codebase.

Although this approach might be fine in the meantime, I would still suggest adopting a more artifact based approach to your 3rd party dependencies.

PR #15, opened 2017-05-22
fix(synchronized): suspected simultaneous access to ArrayList from mu…

Enuviel:
…ltiple threads.

### DIFF
--- a/AndroidSDK/src/com/leanplum/internal/LeanplumManifestHelper.java
+++ b/AndroidSDK/src/com/leanplum/internal/LeanplumManifestHelper.java
@@ -78,7 +78,10 @@ public class LeanplumManifestHelper {
   /**
    * Gets application components from AndroidManifest.xml file.
    */
-  private static void parseManifestNodeChildren() {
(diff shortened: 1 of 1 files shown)


AndroidSDK/src/com/leanplum/internal/LeanplumManifestHelper.java
```diff
@@ -78,7 +78,10 @@
   /**
    * Gets application components from AndroidManifest.xml file.
    */
-  private static void parseManifestNodeChildren() {
+  private static synchronized void parseManifestNodeChildren() {
+    if (manifestData != null) {
```
benmarten: why do we check this, when we overwrite it in line 85 anyway?

Enuviel: Because if manifest is already parsed there is no reason to do it again. It will not change until application is reinstalled.
And it is not immediate operation to parse it because data needs to be pulled from .zip archive

PR #18, opened 2017-06-01
feat(sync) add UUID logic.

Enuviel:


benmarten:
test this please

AndroidSDK/src/com/leanplum/internal/Request.java
```diff
@@ -383,11 +384,13 @@ private void sendNow() {
     Util.executeAsyncTask(true, new AsyncTask<Void, Void, Void>() {
       @Override
       protected Void doInBackground(Void... params) {
-        final List<Map<String, Object>> requestsToSend = getUnsentRequests();
+        List<Map<String, Object>> requestsToSend = getUnsentRequests();
         if (requestsToSend.isEmpty()) {
           return null;
         }
 
+        requestsToSend = updateUnsentRequests(requestsToSend);
```
milos1290: why do we need to update unsent requests ? couldn't we generate uuid when inserting ? This just causes double reading/parsing and saving again.

benmarten: I agree, can the uuid be hold in shared preferences, until next send, and included during inserting of events....
only on sending the events, we will update the uuid?

Enuviel: we should be sure that all requests were saved with generated UUID before sending them to the server.
We will have one UUID for the batch of actions, that means we cannot generate UUID when inserting. We can avoid double parsing with moving UUID logic to getUnsentRequests.

PR #19, opened 2017-06-02
feat(sequentially) add logic around Integer.MAX_VALUE for SharedPrefe…

Enuviel:
…rences key.


benmarten:
Note: This PR will be replaced by SQLlite solution before next release.

PR #27, opened 2017-06-06
fix(inbox): fixing null pointer exception

milos1290:


AndroidSDK/src/com/leanplum/NewsfeedMessage.java
```diff
@@ -149,6 +149,9 @@ public Date deliveryTimestamp() {
    * Returns the delivery timestamp of the newsfeed message.
    */
   public Date getDeliveryTimestamp() {
+    if (deliveryTimestamp == null) {
+      return new Date();
```
Enuviel: I think better return here null, not the current date.

benmarten: - yes, but then guard against null here: https://github.com/Leanplum/Leanplum-Android-SDK/pull/27/files#diff-f7185a5c0a8b80ab803fcb5bd4bf2c50R321
and in all other invocations of this method.
- also we should update docs, to document that it might return null.

milos1290: better return any date then null, we should definitely document it

Enuviel: deliveryTimestamp should be real one or null if not present

AndroidSDK/src/com/leanplum/LeanplumInbox.java
```diff
@@ -306,9 +306,26 @@ public int unreadCount() {
...
       Collections.sort(messageIds, new Comparator<String>() {
         @Override
-        public int compare(String firstMessage, String secondMessage) {
-          Date firstDate = messageForId(firstMessage).getDeliveryTimestamp();
-          Date secondDate = messageForId(secondMessage).getDeliveryTimestamp();
+        public int compare(String firstMessageId, String secondMessageId) {
+          LeanplumInboxMessage firstMessage = messageForId(firstMessageId);
+          LeanplumInboxMessage secondMessage = messageForId(secondMessageId);
+          // Message that is null will be moved to the back of the list.
+          if (firstMessage == null) {
```
Enuviel: move to right after firstMessage init?

AndroidSDK/src/com/leanplum/LeanplumInbox.java
```diff
@@ -306,9 +306,26 @@ public int unreadCount() {
...
+            return -1;
+          }
+          if (secondMessage == null) {
+            return 1;
+          }
+
+          Date firstDate = firstMessage.getDeliveryTimestamp();
+          Date secondDate = secondMessage.getDeliveryTimestamp();
+          // Message with null date will be moved to the back of the list.
+          if (firstDate == null) {
```
Enuviel: move to right after firstDate init?

AndroidSDK/src/com/leanplum/LeanplumInbox.java
```diff
@@ -306,9 +306,26 @@ public int unreadCount() {
...
+          LeanplumInboxMessage secondMessage = messageForId(secondMessageId);
+          // Message that is null will be moved to the back of the list.
+          if (firstMessage == null) {
+            return -1;
+          }
+          if (secondMessage == null) {
+            return 1;
+          }
+
+          Date firstDate = firstMessage.getDeliveryTimestamp();
```
Enuviel: may be move this one  to right after check firstMessage for null?

PR #37, opened 2017-06-16
fix(NPE) wrapping creation notification in try{}catch block to avoid …

Enuviel:
…NPE.

getProfiles at this line
https://github.com/android/platform_frameworks_base/blob/d7b1f41109abee60ec8529369a85d84300b1d678/core/java/android/app/ApplicationPackageManager.java#L2306 can return null and it will crash with NPE at next line.


benmarten:
thx. for these easy review, one reviewer is sufficient imo ;)

PR #41, opened 2017-06-19
fix(HTML-banner) fix for UI elements covered by HTML Banner are click…

Enuviel:
…able through the banner.

### DIFF
--- a/AndroidSDK/src/com/leanplum/utils/SizeUtil.java
+++ b/AndroidSDK/src/com/leanplum/utils/SizeUtil.java
@@ -114,4 +114,17 @@ public class SizeUtil {
     }
     return result;
   }
+
+  public static int getStatusBarHeight(Context context) {
+    init(context);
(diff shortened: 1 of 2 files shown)


AndroidSDK/src/com/leanplum/utils/SizeUtil.java
```diff
@@ -114,4 +114,17 @@ public static int getStatusBarHeight(Activity activity) {
     }
     return result;
   }
+
+  public static int getStatusBarHeight(Context context) {
+    init(context);
+    int result = 0;
+    try {
+      int resourceId = context.getResources().getIdentifier("status_bar_height", "dimen", "android");
```
milos1290: can we rely on this in 7.0 ?

Enuviel: just tested on  7.1.2 - it works

PR #43, opened 2017-06-22
perf(push): Send stored push token as a parameter of start call and w…

Enuviel:
…hen we got a new token.

### DIFF
--- a/AndroidSDK/src/com/leanplum/Leanplum.java
+++ b/AndroidSDK/src/com/leanplum/Leanplum.java
@@ -52,6 +52,7 @@ import com.leanplum.internal.Util;
 import com.leanplum.internal.Util.DeviceIdInfo;
 import com.leanplum.internal.VarCache;
 import com.leanplum.messagetemplates.MessageTemplates;
+import com.leanplum.utils.SharedPreferencesUtil;
 
 import org.json.JSONArray;
 import org.json.JSONObject;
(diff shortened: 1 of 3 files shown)


AndroidSDK/src/com/leanplum/LeanplumCloudMessagingProvider.java
```diff
@@ -75,9 +75,8 @@ void onRegistrationIdReceived(Context context, String registrationId) {
         context, Constants.Defaults.LEANPLUM_PUSH, Constants.Defaults.PROPERTY_REGISTRATION_ID))) {
       Log.i("Device registered for push notifications with registration token", registrationId);
       storePreferences(context.getApplicationContext());
+      sendRegistrationIdToBackend(LeanplumCloudMessagingProvider.registrationId);
```
benmarten: add hasStarted check: https://github.com/Leanplum/Leanplum-iOS-SDK/pull/39/files#diff-e8153ae4bdb7fef57acf322b955d6c15R87

AndroidSDK/src/com/leanplum/Leanplum.java
```diff
@@ -628,6 +629,9 @@ private static void startHelper(
     Date now = new Date();
     int timezoneOffsetSeconds = localTimeZone.getOffset(now.getTime()) / 1000;
 
+    String registrationId = SharedPreferencesUtil.getString(context,
```
benmarten: looks like in iOS: we also include appId etc.: https://github.com/Leanplum/Leanplum-iOS-SDK/pull/39/files#diff-37d7318a53e27e0c036850d334aace01R427
do we need this for android?

Enuviel: this not needed, before send request we will attach appId etc: https://github.com/Leanplum/Leanplum-Android-SDK/blob/develop/AndroidSDK/src/com/leanplum/internal/Request.java#L293

plus when you createArgsDictionary  - you added deviceID and userID
https://github.com/Leanplum/Leanplum-Android-SDK/blob/develop/AndroidSDK/src/com/leanplum/internal/Request.java#L185

AndroidSDK/src/com/leanplum/LeanplumCloudMessagingProvider.java
Enuviel: we don't need this check.

PR #45, opened 2017-06-29
fix(ManifestHelper) fix close JarFile.

Enuviel:
We forgot to close JarFile. JarFile is a closable object. Client has enabled StrictMode with detectLeakedClosableObjects(). And reported it.


benmarten:
can you please provide a little context in the description? is this from exception reports?

PR #49, opened 2017-07-04
feat(tests): opensourcing tests

milos1290:


milos1290:
@benmarten no modification is made on source files, only test module is added.

Enuviel:
Thanks!!!

PR #54, opened 2017-07-06
fix(Callbacks) Handle callbacks for event if here more then 10k events.

Enuviel:
There can be a corner case when the last response/error callback is not the response/error we are looking for.

### DIFF
--- a/AndroidSDK/src/com/leanplum/Leanplum.java
+++ b/AndroidSDK/src/com/leanplum/Leanplum.java
@@ -666,11 +666,11 @@ public class Leanplum {
     Util.initializePreLeanplumInstall(params);
 
     // Issue start API call.
-    Request req = Request.post(Constants.Methods.START, params);
+    final Request req = Request.post(Constants.Methods.START, params);
     req.onApiResponse(new Request.ApiResponseCallback() {
       @Override
       public void response(List<Map<String, Object>> requests, JSONObject response) {
-        Leanplum.handleApiResponse(response, requests);
+        Leanplum.handleApiResponse(response, requests, req);
       }
     });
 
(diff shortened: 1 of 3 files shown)


AndroidSDK/src/com/leanplum/internal/LeanplumEventCallbackManager.java
```diff
@@ -0,0 +1,126 @@
...
+import android.support.annotation.NonNull;
+
+import org.json.JSONObject;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * LeanplumEventCallbackManager class for handle events callbacks.
```
benmarten: `...class *to* handle *event*...`

AndroidSDK/src/com/leanplum/internal/Request.java
```diff
@@ -110,6 +114,14 @@ public static String token() {
     return token;
   }
 
+  public long getDataBaseIndex() {
```
benmarten: javadoc to describe that the databaseIndex is used for

AndroidSDK/src/com/leanplum/Leanplum.java
```diff
@@ -706,176 +712,184 @@ private static void handleApiResponse(JSONObject response, List<Map<String, Obje
...
   }
 
-  private static void handleStartResponse(JSONObject response) {
-    boolean success = Request.isResponseSuccess(response);
-    if (!success) {
-      try {
-        LeanplumInternal.setHasStarted(true);
-        LeanplumInternal.setStartSuccessful(false);
+  private static void handleStartResponse(final JSONObject response) {
+    Util.executeAsyncTask(false, new AsyncTask<Void, Void, Void>() {
```
Enuviel: we need to handle Start response in the different thread, at this time synchronized block will work correctly, when we will call getNewsfeedMessages.

PR #65, opened 2017-07-26
fix(networking): fix LP-5245  - start ResponseCallback from different…

Enuviel:
… thread.

### DIFF
--- a/AndroidSDK/src/com/leanplum/internal/LeanplumEventCallbackManager.java
+++ b/AndroidSDK/src/com/leanplum/internal/LeanplumEventCallbackManager.java
@@ -75,6 +75,9 @@ class LeanplumEventCallbackManager {
     // Loop over all callbacks.
     for (; iterator.hasNext(); ) {
       final Map.Entry<Request, LeanplumEventCallbacks> entry = iterator.next();
+      if (entry.getKey() == null) {
+        continue;
+      }
       if (entry.getKey().getDataBaseIndex() >= countOfEvents) {
         entry.getKey().setDataBaseIndex(entry.getKey().getDataBaseIndex() - countOfEvents);
       } else {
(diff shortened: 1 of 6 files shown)


AndroidSDK/src/com/leanplum/internal/LeanplumEventCallbackManager.java
```diff
@@ -101,12 +110,21 @@ void invokeAllCallbacksForResponse(@NonNull JSONObject responseBody, int countOf
...
         entry.getKey().setDataBaseIndex(entry.getKey().getDataBaseIndex() - countOfEvents);
       } else {
         if (entry.getValue() != null && entry.getValue().responseCallback != null) {
-          entry.getValue().responseCallback.response(responseBody);
+          // Start callback asynchronously, to avoid creation of new Request object from the same
+          // thread.
+          Util.executeAsyncTask(false, new AsyncTask<Void, Void, Void>() {
+            @Override
+            protected Void doInBackground(Void... params) {
+              entry.getValue().responseCallback.response(Request.getResponseAt(responseBody,
```
benmarten: check for null pointer here?

AndroidSDK/src/com/leanplum/internal/LeanplumEventCallbackManager.java
```diff
@@ -73,12 +74,20 @@ void invokeAllCallbacksWithError(@NonNull Exception error, int countOfEvents) {
...
         entry.getKey().setDataBaseIndex(entry.getKey().getDataBaseIndex() - countOfEvents);
       } else {
         if (entry.getValue() != null && entry.getValue().errorCallback != null) {
-          entry.getValue().errorCallback.error(error);
+          // Start callback asynchronously, to avoid creation of new Request object from the same
+          // thread.
+          Util.executeAsyncTask(false, new AsyncTask<Void, Void, Void>() {
+            @Override
+            protected Void doInBackground(Void... params) {
+              entry.getValue().errorCallback.error(error);
```
benmarten: do we need to guard for null pointers here maybe?

PR #68, opened 2017-08-03
Adding track method with argument parameter.

richardortiz84:
Adds a new track method which allows a user to send in arguments to a leanplum api call such as currency codes.

PR #69, opened 2017-08-07
Adding track method with argument parameter.

benmarten:
This is the #68 with added unit test.

### DIFF
--- a/AndroidSDK/src/com/leanplum/Leanplum.java
+++ b/AndroidSDK/src/com/leanplum/Leanplum.java
@@ -1541,6 +1541,30 @@ public class Leanplum {
   }
 
   /**
+   * Logs a particular event in your application. The string can be any value of your choosing, and
+   * will show up in the dashboard.
+   * <p>
+   * <p>To track Purchase events, call {@link Leanplum#trackGooglePlayPurchase} instead for in-app
+   * purchases, or use {@link Leanplum#PURCHASE_EVENT_NAME} as the event name for other types of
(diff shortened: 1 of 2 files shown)


benmarten:
@richardortiz84 We discussed this PR further in the team, and we are hesitant to expose the args parameter completely, as this could lead to confusion among other customers.
I believe what you were looking for could be achieved by using https://github.com/Leanplum/Leanplum-Android-SDK/blob/master/AndroidSDK/src/com/leanplum/Leanplum.java#L1528 ?
I don't think currencyCode is available in iOS but we can add a similar method for iOS....

richardortiz84:
@benmarten Calling another event does not achieve what we are trying to do. We have another event that is being used for track purchases where we send in a currency. The track function not having a currency code is a problem because our application takes in currencies other than USD and we need those converted so that we can properly target our audience.

alexisoyama:
Hi @richardortiz84. I believe what you are looking for is a track method that takes currency code. This is a valid case for non in-app purchase apps. I need to confirm with the team but we will most likely make a new method that will cover this case. If so we will make it. No need to make a PR.

benmarten:
I'm not sure if we made any progress on this request. Let's discuss in todays sprint planning...

benmarten:
We have added a task in our current sprint, to add a new method to support this use case.

benmarten:
Resolved by https://github.com/Leanplum/Leanplum-Android-SDK/pull/77 @richardortiz84

PR #76, opened 2017-08-25
feat(service): fixing a crash if service is started from broadcast re…

milos1290:
…ceiver

### DIFF
--- a/AndroidSDK/src/com/leanplum/LeanplumPushService.java
+++ b/AndroidSDK/src/com/leanplum/LeanplumPushService.java
@@ -657,6 +657,7 @@ public class LeanplumPushService {
     } catch (LeanplumException e) {
       Log.e("There was an error registering for push notifications.\n" +
           Log.getStackTraceString(e));
+    } catch (Throwable ignored) {
     }
   }
 


AndroidSDK/src/com/leanplum/LeanplumPushService.java
```diff
@@ -657,6 +657,7 @@ static void onStart() {
     } catch (LeanplumException e) {
       Log.e("There was an error registering for push notifications.\n" +
           Log.getStackTraceString(e));
+    } catch (Throwable ignored) {
```
benmarten: Why is this throwable ignored? Should we log it also, so customer or at least we know that there's an issue?

milos1290: There is an error log when SDK is started from Application.onCreate() on Android O due to push notification broadcast receiver, which will try to start the sdk and service to register token, and due to new background policies it will fail since we are not allowed to start service in background. The app isn't actually started, the process is only created because of broadcast receiver, this fix basically suppresses that log to be shown in console.

Enuviel: for me it still not safe, we should discuss this PR offline.

benmarten: I think Anna tested it, and it seems that messages are lost when the device reboots. We should not have any dataloss, as this would affect customer negatively. I propose we spec out the alternative solution. What do  you think?

benmarten:
https://leanplum.atlassian.net/browse/LP-5371

AndroidSDK/src/com/leanplum/LeanplumPushService.java
milos1290: Isn't that default behavior on Android ? The notifications won't be received after reboot until app is started at least once ?

benmarten: Ok, so further testing has shown, that messages will not be lost, but delivered after a couple of minutes by gcm. so we should be fine with this fix for now.

PR #77, opened 2017-08-30
feat(api): adding purchase track method to manually track purchases o…

milos1290:
…utside IAPs.

### DIFF
--- a/AndroidSDK/src/com/leanplum/Leanplum.java
+++ b/AndroidSDK/src/com/leanplum/Leanplum.java
@@ -1576,6 +1576,7 @@ public class Leanplum {
 
       LeanplumInternal.track(event, value, null, params, requestArgs);
     } catch (Throwable t) {
+      Log.e("trackPurchase - Failed to track purchase event.");
       Util.handleException(t);
     }
   }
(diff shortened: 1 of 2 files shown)


milos1290:
Done.

AndroidSDK/src/com/leanplum/Leanplum.java
```diff
@@ -1553,6 +1553,34 @@ public static void track(final String event, double value, String info,
...
+      }
+
+      final Map<String, String> requestArgs = new HashMap<>();
+      if (!TextUtils.isEmpty(currencyCode)) {
+        requestArgs.put(Constants.Params.IAP_CURRENCY_CODE, currencyCode);
+      }
+
+      LeanplumInternal.track(event, value, null, params, requestArgs);
+    } catch (Throwable t) {
+      Util.handleException(t);
```
Enuviel: May be to put a Log?

benmarten: Where?

Enuviel: before handleException , something like Log.e("trackPurchase - Failed to track purchase event.");

PR #79, opened 2017-08-31
feat(channel): fixing json parsing issue

milos1290:


AndroidSDK/src/com/leanplum/LeanplumNotificationChannel.java
```diff
@@ -602,10 +604,14 @@ private static int getTargetSdkVersion(Context context) {
       showBadge = (boolean) CollectionUtil.getOrDefault(channel, "show_badge", showBadge);
 
       try {
-        List<Long> pattern = CollectionUtil.uncheckedCast(
+        List<Number> pattern = CollectionUtil.uncheckedCast(
             CollectionUtil.getOrDefault(channel, "vibration_pattern", null));
```
benmarten: Are you sure this will work?
in case of default value wouldn't it be:
`List<Number> pattern = (List<Number>)"vibrationPattern"` which is trying to cast a string to a list?

AndroidSDK/src/com/leanplum/LeanplumNotificationChannel.java
```diff
@@ -602,10 +604,14 @@ private static int getTargetSdkVersion(Context context) {
...
         if (pattern != null) {
-          vibrationPattern = CollectionUtil.toPrimitive(pattern.toArray(new Long[pattern.size()]));
+          vibrationPattern = new long[pattern.size()];
+          Iterator<Number> iterator = pattern.iterator();
+          for (int i = 0; i < vibrationPattern.length; i++) {
```
benmarten: is iterator better than using foreach?

AndroidSDK/src/com/leanplum/LeanplumNotificationChannel.java
```diff
@@ -602,10 +604,14 @@ private static int getTargetSdkVersion(Context context) {
...
+            vibrationPattern[i] = iterator.next().longValue();
```
benmarten: npe check?

PR #84, opened 2017-09-06
feat(deprecation): Revert deprecation  for count and numberValue meth…

Enuviel:
Remove deprecation for count and numberValue methods at Var.java.
Looks like these methods was deprecated by mistake.


AndroidSDK/src/com/leanplum/Var.java
```diff
@@ -547,17 +537,7 @@ private int countInternal() {
    *
    * @return A Number value.
```
benmarten: Should we reuse the other javadoc here? `Gets a value from a variable initialized as Number.`

benmarten:
This should actually have a `revert` commit title!

PR #101, opened 2017-10-04
fix: Retrieve chained message if it doesn't exist locally yet.

lan17:
https://leanplum.atlassian.net/browse/LP-6124?filter=-1

Tested in staging by setting up a case where:

  * App is already running.
  * On dashboard I create a new In-App message that is active.
  * Create a push message with open existing message pointing to the just created in-app message.
  * Send out Push as preview, and have the test app display in-app message after i open the push notification.

### DIFF
--- a/AndroidSDK/src/com/leanplum/ActionContext.java
+++ b/AndroidSDK/src/com/leanplum/ActionContext.java
@@ -374,8 +374,8 @@ public class ActionContext extends BaseActionContext implements Comparable<Actio
   /**
    * Return true if here was an action for this message and we started it.
    */
-  private boolean createActionContextForMessageId(String messageAction, Map<String, Object>
-      messageArgs, String messageId, String name, Boolean chained) {
+  private boolean createActionContextForMessageId(String messageAction,
+      Map<String, Object> messageArgs, String messageId, String name, Boolean chained) {
     try {
       final ActionContext actionContext = new ActionContext(messageAction,
           messageArgs, messageId);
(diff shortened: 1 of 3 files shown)


Enuviel:
test this please

AndroidSDK/src/com/leanplum/LeanplumPushService.java
```diff
@@ -383,7 +383,33 @@ private static void showNotification(Context context, Bundle message) {
     }
 
     try {
-      notificationManager.notify(notificationId, builder.build());
+      // Check if we have a chained message, and if it exists in var cache.
```
Enuviel: could you please create a method for this?

lan17: Where should it go?  Same file?

Enuviel: ActionContext.java ?

lan17: The logic here works on `Bundle` and ActionContext.java deals with `Map`.  Would that still be appropriate?

lan17: I extracted teh logic that works on just the action map to ActionContext.

AndroidSDK/src/com/leanplum/ActionContext.java
```diff
@@ -406,27 +406,55 @@ public void variablesChanged() {
...
+        });
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Extract chained messageId from parent message's actionMap.  If it doesn't exist then return null.
+   */
+  public static String getChainedMessageId(Map<String, Object> actionMap) {
```
Enuviel: please remove public

AndroidSDK/src/com/leanplum/LeanplumPushService.java
Enuviel: you can pass String  to this method - String actionData = message.getString(Constants.Keys.PUSH_MESSAGE_ACTION) and return true/false

lan17:
Added unit tests and moved things around :)

AndroidSDK/src/com/leanplum/LeanplumPushService.java
lan17: fixed

lan17: fixed

PR #103, opened 2017-10-10
feat(tests): adding test coverage reports

milos1290:


AndroidSDK/src/main/java/com/leanplum/LeanplumPushService.java
```diff
@@ -35,6 +35,7 @@
 import android.os.Bundle;
 import android.support.v4.app.NotificationCompat;
 import android.text.TextUtils;
+import android.widget.RemoteViews;
```
Enuviel: I think this one not from this code change.

AndroidSDK/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">AndroidSDK</string>
```
Enuviel: do we use this one only for tests?

milos1290: we don't actually use it anywhere.

AndroidSDK/src/main/java/com/leanplum/LeanplumPushService.java
milos1290: yeah it looks like. no idea how it got through.

Enuviel:
test this please

Enuviel:
Blocked by https://github.com/powermock/powermock/issues/727

PR #125, opened 2017-12-22
fix(html): Update bridge URL.

Enuviel:
There were JavaScript changes in Chrome 63 that broken our HTML messages.

### DIFF
--- a/AndroidSDK/src/com/leanplum/messagetemplates/MessageTemplates.java
+++ b/AndroidSDK/src/com/leanplum/messagetemplates/MessageTemplates.java
@@ -57,6 +57,7 @@ public class MessageTemplates {
     static final String LAYOUT_HEIGHT = "Layout.Height";
     static final String HTML_WIDTH = "HTML Width";
     static final String HTML_HEIGHT = "HTML Height";
+    static final String HTML_USE_NEW_BASE_URL = "useNewBaseUrl";
     static final String HTML_Y_OFFSET = "HTML Y Offset";
     static final String HTML_TAP_OUTSIDE_TO_CLOSE = "Tap Outside to Close";
     static final String HTML_ALIGN = "HTML Align";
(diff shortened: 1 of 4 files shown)


Enuviel:
test this please

AndroidSDK/src/com/leanplum/messagetemplates/MessageTemplates.java
```diff
@@ -91,17 +92,17 @@
     static final String DEFAULT_URL = "http://www.example.com";
 
     // Web interstitial values.
-    static final String DEFAULT_CLOSE_URL = "http://leanplum:close";
+    static final String DEFAULT_CLOSE_URL = "https://leanplum/close";
```
acconway: Can you pull the base URL out into it's own value, so in the future we don't have to change multiple values?

```
DEFAULT_BASE_URL = "https://leanplum/"
static final String DEFAULT_CLOSE_URL = DEFAULT_BASE_URL + "close";
```

AndroidSDK/src/com/leanplum/messagetemplates/MessageTemplates.java
```diff
@@ -91,17 +92,17 @@
...
 
     // HTML Template values.
     public static final String FILE_PREFIX = "__file__";
     public static final String HTML_TEMPLATE_PREFIX = "__file__Template";
-    static final String DEFAULT_OPEN_URL = "http://leanplum:loadFinished";
-    static final String DEFAULT_TRACK_URL = "http://leanplum:track";
-    static final String DEFAULT_ACTION_URL = "http://leanplum:runAction";
-    static final String DEFAULT_TRACK_ACTION_URL = "http://leanplum:runTrackedAction";
-
+    static final String DEFAULT_OPEN_URL = "https://leanplum/loadFinished";
```
acconway: I think it would be safer to handle both the old and new URLs. It shouldn't be possible but just in case they got the new SDK with an old template we should cover that case.

AndroidSDK/src/com/leanplum/messagetemplates/MessageTemplates.java
```diff
@@ -90,18 +90,19 @@
 
     // Open URL.
     static final String DEFAULT_URL = "http://www.example.com";
-
+    static final String DEFAULT_BASE_URL_NEW = "https://leanplum/";
```
alexisoyama: I don't think we need to have both New and Old one. Can we make it an ActionContext argument and have "https://leanplum/" as the default one?

acconway: IMO it's safer to have both. Just in case there is a bug somewhere else in the system, and maybe the new template isn't downloaded.

acconway:
@alexisoyama This is the quickest fix which is that SDK just sets a boolean value for template whether to use old or new URL

acconway:
@alexisoyama I created a follow up ticket to send more information from SDK about version, etc, but we should spend more time figuring out the format for that, something that works for both platforms

AndroidSDK/src/com/leanplum/messagetemplates/MessageTemplates.java
alexisoyama: In that case we should handle that in the bridge and not pollute the ActionContext arguments.

alexisoyama:
@acconway @Enuviel 
So this worked on my end. https://phabricator.leanplum.com/D4281
When you guys have a chance, can you test it on old Android? The payload should be the same, but just to make sure...

`lpm android:configure --app-id=app_Lqrqzgorj2NKrchJ4c3QeX2hhrH97Qx4GYS3wfHbLyI --prod-key=prod_anRZRibrDXHQHYFwshVXbW2Em7zNdMTWWEmhqqDHaa4 --dev-key=dev_xGJU1n6UzR1N8FyRXffIh0mstNYmWv1trGZOsQPctQE --environment=staging
`

PR #126, opened 2017-12-22
fix(services):  Adds temporary try/catch to avoid crashes.

Enuviel:


Enuviel:
test this please

Enuviel:
We don't need to log something.

Enuviel:
We don't need to log something.

alexisoyama:
Okay, that’s fine but can you tell me more about the fix / what’s going on here?

Enuviel:
test this please

sayaansaha:
test this please

PR #150, opened 2018-01-26
feat(push): Checks if client compiles leanplum-gcm and leanplum-fcm at…

Enuviel:
feat(push): Checks if client compiles leanplum-gcm and leanplum-fcm at the same time.


Enuviel:
test this please

PR #171, opened 2018-04-19
Custom locale configuration in start process and runtime

rubenortiz-IM:
## Proposed Changes

- added new builder to simplify start parameters (including new locale configuration)
- added new custom locale parameter for the start process
- added new start method to support the new parameter builder
- generic start supports custom locale (null value for locale has the previous behaviour)
- add setter for the locale and propagate the changes to the server (via START call)
- added unit test for the builder and the new locale parameter
- added unit test for the locale setter


rubenortiz-IM:
why closing the pull request? Do you have a reason to don't support custom locales?

PR #174, opened 2018-04-26
feat(Networking): Reduce MAX_EVENTS_PER_API_CALL to 1000 and send req…

Enuviel:
feat(Networking): Reduce MAX_EVENTS_PER_API_CALL to 1000 and send a request to the server on 1000 events in the database.

## Proposed Changes

  - Reduce MAX_EVENTS_PER_API_CALL from 10000 to 1000.
  - Send a request to the server on 1000 events in the database.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/internal/JsonConverter.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/internal/JsonConverter.java
@@ -45,7 +45,7 @@ public class JsonConverter {
     }
     try {
       return mapToJsonObject(map).toString();
-    } catch (JSONException e) {
+    } catch (Throwable e) {
       Log.e("Error converting " + map + " to JSON", e);
       return null;
     }
@@ -57,7 +57,7 @@ public class JsonConverter {
     }
     try {
       return mapFromJson(new JSONObject(json));
-    } catch (JSONException e) {
+    } catch (Throwable e) {
       Log.e("Error converting " + json + " from JSON", e);
       return null;
     }
@@ -141,21 +141,26 @@ public class JsonConverter {
     if (json == null) {
       return null;
     }
-    List<Object> result = new ArrayList<>(json.length());
-    for (int i = 0; i < json.length(); i++) {
-      Object value = json.opt(i);
-      if (value == null || value == JSONObject.NULL) {
-        value = null;
-      } else if (value instanceof JSONObject) {
-        value = mapFromJson((JSONObject) value);
-      } else if (value instanceof JSONArray) {
-        value = listFromJson((JSONArray) value);
-      } else if (JSONObject.NULL.equals(value)) {
-        value = null;
+    try {
+      List<Object> result = new ArrayList<>(json.length());
+      for (int i = 0; i < json.length(); i++) {
+        Object value = json.opt(i);
+        if (value == null || value == JSONObject.NULL) {
+          value = null;
+        } else if (value instanceof JSONObject) {
+          value = mapFromJson((JSONObject) value);
+        } else if (value instanceof JSONArray) {
+          value = listFromJson((JSONArray) value);
+        } else if (JSONObject.NULL.equals(value)) {
+          value = null;
+        }
+        result.add(value);
       }
-      result.add(value);
+      return CollectionUtil.uncheckedCast(result);
+    } catch (Throwable t) {
+      Log.e("Error converting list " + json + " from JSON.", t);
+      return null;
     }
-    return CollectionUtil.uncheckedCast(result);
   }
 
   public static <T> List<T> listFromJsonOrDefault(JSONArray json) {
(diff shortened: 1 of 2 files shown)


AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -461,18 +464,18 @@ private void sendNow() {
...
       protected Void doInBackground(Void... params) {
-        sendRequests();
+        sendRequests(true);
         return null;
       }
     });
   }
 
-  private void sendRequests() {
+  private void sendRequests(boolean allowMoreThanOneNetworkRequest) {
```
afirst: What is the purpose of this argument and can you document it?

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -565,6 +568,10 @@ private void sendRequests() {
   }
 
   public void sendEventually() {
+    sendEventually(false);
+  }
+
+  public void sendEventually(boolean sendNow) {
```
afirst: This argument is confusing. There's already a sendNow method. Methods should do 1 thing, which should be clear from the name of the method.

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -230,7 +221,7 @@ public void onApiResponse(ApiResponseCallback apiResponse) {
     return args;
   }
 
-  private void saveRequestForLater(final Map<String, Object> args) {
+  private void saveRequestForLater(final Map<String, Object> args, final boolean sendNow) {
```
afirst: This method should do 1 thing. If you need to do something afterwards, perhaps you can pass a callback to run additional code when this finishes?

AndroidSDKCore/src/main/java/com/leanplum/internal/JsonConverter.java
```diff
@@ -45,7 +45,7 @@ public static String toJson(Map<String, ?> map) {
     }
     try {
       return mapToJsonObject(map).toString();
-    } catch (JSONException e) {
+    } catch (Throwable e) {
```
afirst: Is it safe to be returning null everywhere? Could we incur data loss or other issues? Would we end up with NPEs instead of OOMs?

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -461,18 +464,18 @@ private void sendNow() {
...
     }
 
-    this.sendEventually();
+    this.sendEventually(true);
 
     Util.executeAsyncTask(true, new AsyncTask<Void, Void, Void>() {
       @Override
       protected Void doInBackground(Void... params) {
-        sendRequests();
+        sendRequests(true);
```
afirst: Can you help me understand this logic? It seems like there'd be a race condition, where in the previous `sendEventually` method, we're going to call `sendRequests` in an async task, and then we do it again here in another async task. Why do we need to do this twice, and does it matter if one potentially runs before the other?

Enuviel: each of sending async tasks is queued into SingleThreadExecutor, so they will be executed sequentially, not in parallel.

AndroidSDKCore/src/main/java/com/leanplum/internal/JsonConverter.java
Enuviel: This method returned null before. Removed Throwable - added only OutOfMemory catch.

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -461,18 +457,30 @@ private void sendNow() {
...
-  private void sendRequests() {
+  /**
+   * Because saveRequestForLater() can send some requests now, we want to send only completed
+   * batches with 1000 events, not incomplete batches with few events.
+   * Due to lack of sendRequests(int limit) method, we use flag to simulate such behavior.
+   *
+   * @param onlyFirstBatch - by default is false. If true - we send only first batch of events and
+   * stop there.
+   */
+  private void sendRequests(boolean onlyFirstBatch) {
```
alexisoyama: I think this naming is confusing. Similar to sendNow(false), when I see sendRequests(false) I’m just confused. Is this sending request or not? You need to read the doc to understand that this flag indicates one time only request. I will make make this is a helper function and have a 2 functions that  calls it.

Enuviel: Sure, feel free to suggest a better version

Enuviel:
I'm happy to replace it with simple code that doesn't require deep thinking. Though I don't see any quick solution for that. Is there something specific you are talking about?

alexisoyama:
I’m just confused by why we need have a callback that sends another request on top of what we have now. Isn’t the goal to change the limit from 10k to 1k with additional catching?

alexisoyama:
Synced offline but want to make it clear. We cannot add extra layer of functionality until refactoring is done. Sending additional request creates more risk that even our functional test might not cover. It’s a high risk code and needs to be treated carefully.

PR #176, opened 2018-05-16
Wrap doInBackgrond with try/catch where missing

e7mac:

PR #182, opened 2018-05-18
Back off 50%

e7mac:


AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -480,22 +480,83 @@ protected Void doInBackground(Void... params) {
...
+    List<Map<String, Object>> requestsToSend;
+    String jsonEncodedString;
+  }
+
+  private RequestsWithEncoding getRequestsWithEncodedStringForErrors() {
+
     List<Map<String, Object>> unsentRequests = new ArrayList<>();
     List<Map<String, Object>> requestsToSend;
+    String jsonEncodedRequestsToSend;
+
```
mohammedsameer: Minor: Extra line

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -480,22 +480,83 @@ protected Void doInBackground(Void... params) {
...
+    RequestsWithEncoding requestsWithEncoding = new RequestsWithEncoding();
+    requestsWithEncoding.unsentRequests = unsentRequests;
+    requestsWithEncoding.requestsToSend = requestsToSend;
+    requestsWithEncoding.jsonEncodedString = jsonEncodedRequestsToSend;
+
+    return requestsWithEncoding;
+  }
+
+
+  protected RequestsWithEncoding getRequestsWithEncodedStringStoredRequests(double fraction) {
```
mohammedsameer: Minor: Javadoc

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -480,22 +480,83 @@ protected Void doInBackground(Void... params) {
...
+      }
+
+      jsonEncodedRequestsToSend = jsonEncodeUnsentRequests(unsentRequests);
+
+      requestsWithEncoding.unsentRequests = unsentRequests;
+      requestsWithEncoding.requestsToSend = requestsToSend;
+      requestsWithEncoding.jsonEncodedString = jsonEncodedRequestsToSend;
+
+      return requestsWithEncoding;
+    } catch (OutOfMemoryError E) {
```
mohammedsameer: Maybe a comment within the catch block suggesting why we are catching OOO and still returning requests with 50% multiplier

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
```diff
@@ -398,14 +398,14 @@ public void testCrashes() throws Exception {
...
+    Request request2 = new Request("POST", Constants.Methods.LOG, null);
 
+    request1.sendEventually();
+    request2.sendEventually();
     // Get a number of events in the database.
     // Expectation: 2 events.
-    Method getUnsentRequests = Request.class.getDeclaredMethod("getUnsentRequests");
-    getUnsentRequests.setAccessible(true);
-    List unsentRequests = (List) getUnsentRequests.invoke(Request.class);
+    List unsentRequests = request1.getUnsentRequests(1.0);
```
mohammedsameer: Minor: Declare a constant FRACTION=1.0. Looks like its being reused few times.

AndroidSDKTests/src/test/java/com/leanplum/internal/RequestTest.java
```diff
@@ -198,12 +202,112 @@ public void testRemoveIrrelevantBackgroundStartRequests() throws NoSuchMethodExc
...
+
+  @Test
+  public void testJsonEncodeUnsentRequestsWithExceptionLargeNumbers() throws NoSuchMethodException,
+          InvocationTargetException, IllegalAccessException {
+    LeanplumEventDataManager.init(Leanplum.getContext());
+    Request.RequestsWithEncoding requestsWithEncoding;
+    // Prepare testable objects and method.
+    Request request = spy(new Request("POST", Constants.Methods.START, null));
+    request.sendEventually();
+    for (int i = 0;i < 4999; i++) {
```
mohammedsameer: Minor: Space between 0; i <

AndroidSDKTests/src/test/java/com/leanplum/internal/RequestTest.java
```diff
@@ -198,12 +202,112 @@ public void testRemoveIrrelevantBackgroundStartRequests() throws NoSuchMethodExc
...
+      new Request("POST", Constants.Methods.START, null).sendEventually();
+    }
+    // Expectation: 5000 requests returned.
```
mohammedsameer: Isn't it 4999 ? Because loop runs from 0 to < 4999 and not 0 to <= 4999

AndroidSDKTests/src/test/java/com/leanplum/internal/RequestTest.java
```diff
@@ -198,12 +202,112 @@ public void testRemoveIrrelevantBackgroundStartRequests() throws NoSuchMethodExc
...
+    assertEquals(2500, requestsWithEncoding.unsentRequests.size());
+
+    // Throw OOM on 2500, 5000 requests
+    // Expectation: 1250 requests returned.
+    when(request.getUnsentRequests(0.5)).thenThrow(OutOfMemoryError.class);
+    requestsWithEncoding = request.getRequestsWithEncodedStringStoredRequests(1.0);
+    assertEquals(1250, requestsWithEncoding.unsentRequests.size());
+
+    // Throw OOM on > 0 requests
+    // Expectation: 0 requests returned but no crash
```
mohammedsameer: If I am understanding right 50% of requests are returned during OOM but don't crash

AndroidSDKTests/src/test/java/com/leanplum/internal/RequestTest.java
```diff
@@ -198,12 +202,112 @@ public void testRemoveIrrelevantBackgroundStartRequests() throws NoSuchMethodExc
...
+
+    Request realRequest = new Request("POST", Constants.Methods.START, null);
+    Request request = spy(realRequest);
+    when(request.getUnsentRequests(1.0)).thenReturn(requests);
+
+    Request.RequestsWithEncoding requestsWithEncoding = request.getRequestsWithEncodedStringStoredRequests(1.0);
+
+    assertEquals("{\"data\":[{\"0\":\"0\"},{\"0\":\"1\"},{\"1\":\"0\"},{\"1\":\"1\"}]}", requestsWithEncoding.jsonEncodedString);
+  }
+  @Test
```
mohammedsameer: Minor: Add extra space

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -480,22 +480,88 @@ protected Void doInBackground(Void... params) {
...
+
+
+  protected RequestsWithEncoding getRequestsWithEncodedStringStoredRequests(double fraction) {
+    try {
+      List<Map<String, Object>> unsentRequests;
+      List<Map<String, Object>> requestsToSend;
+      String jsonEncodedRequestsToSend;
+      RequestsWithEncoding requestsWithEncoding = new RequestsWithEncoding();
+
+      if (fraction < 0.01) { //base case
```
divyas1102: @e7mac Is there a reason, we are dropping the request after ~6 attempts i.e around 100 requests?
Could we add some server-side logging or at least a metric here so we know when this happens?

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -480,22 +480,88 @@ protected Void doInBackground(Void... params) {
...
     List<Map<String, Object>> requestsToSend;
+    String jsonEncodedRequestsToSend;
+
+    String uuid = UUID.randomUUID().toString();
+    for (Map<String, Object> error : localErrors) {
+      error.put(UUID_KEY, uuid);
+      unsentRequests.add(error);
+    }
+    requestsToSend = unsentRequests;
+    jsonEncodedRequestsToSend = jsonEncodeUnsentRequests(unsentRequests);
```
divyas1102: I understand this is only for errors, but isn't there a risk that this `unsentRequests` also grows too big and crashes when `JsonConverter.toJson(data);` is called in line 726?

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -480,22 +480,88 @@ protected Void doInBackground(Void... params) {
...
+
+      jsonEncodedRequestsToSend = jsonEncodeUnsentRequests(unsentRequests);
+      requestsWithEncoding.unsentRequests = unsentRequests;
+      requestsWithEncoding.requestsToSend = requestsToSend;
+      requestsWithEncoding.jsonEncodedString = jsonEncodedRequestsToSend;
+
+      return requestsWithEncoding;
+    } catch (OutOfMemoryError E) {
+      // half the requests will need less memory, recursively
+      return getRequestsWithEncodedStringStoredRequests(0.5 * fraction);
```
divyas1102: @e7mac Could we please add logs for this also, so we can get closer to identifying root cause of the OOM?

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
divyas1102: @e7mac Also, when it hits this code, it means nothing succeeded so far. Shouldn't we clear the entire existing requests in `LeanplumEventDataManager.getEvents` otherwise there will be another attempt to send all these requests the next time?

e7mac: @divyas1102 We have server side logging and we will be monitoring that closely as we roll out the long-term fix.
We are looking for a solution without data loss. The current behavior of crashing is unacceptable and is being addressed here. Beyond that, we want to preserve the data and try sending it later when network and/or memory conditions might be better.

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
e7mac: Indeed. This whole process is wrapped in a try/catch block as in #176 so if errors is too large, we won't crash. Granted, we are not backing off there, but the error payload is very unlikely to be large. Event payload can grow since the param dictionary is unbounded.

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
e7mac: Sure. I'll add that in for the next iteration where we address the root cause. We have already started QAing this build to help you mitigate the crashes ASAP.

PR #190, opened 2018-06-19
Add debug variant info

e7mac:


lan17:
Someone from your own pod should definitely review this as well.

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
```diff
@@ -322,6 +323,14 @@ public static boolean hasReceivedDiffs() {
     return hasReceivedDiffs;
   }
 
+  public static Map<String, Object> getVariantDebugInfo() {
```
lan17: Create and use a class to represent this info such as `VariantDebugInfo` which could contain a collection of `AbTestVariableOverride` or something similar.

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
```diff
@@ -322,6 +323,14 @@ public static boolean hasReceivedDiffs() {
...
+    return variantDebugInfo;
+  }
+
+  public static void setVariantDebugInfo(Map<String, Object> _variantDebugInfo) {
```
lan17: No need for `_` in front of var.

Simply do `this.variantDebugInfo= _variantDebugInfo;`

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
```diff
@@ -322,6 +323,14 @@ public static boolean hasReceivedDiffs() {
...
+    variantDebugInfo= _variantDebugInfo;
```
lan17: Auto format according to our code style, please.

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
```diff
@@ -87,6 +87,7 @@
   private static boolean silent;
   private static int contentVersion;
   private static Map<String, Object> userAttributes;
+  private static Map<String, Object> variantDebugInfo;
```
lan17: Similar comment elsewhere, but this should be its own class, not just a map.

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumInternal.java
```diff
@@ -673,6 +674,14 @@ public static boolean getIsScreenTrackingEnabled() {
     return isScreenTrackingEnabled;
   }
 
+  public static boolean getIsVariantDebugInfoEnabled() {
+    return isVariantDebugInfoEnabled;
+  }
+
+  public static void setIsVariantDebugInfoEnabled(boolean _isVariantDebugInfoEnabled) {
+    isVariantDebugInfoEnabled= _isVariantDebugInfoEnabled;
```
lan17: `this. isVariantDebugInfoEnabled= isVariantDebugInfoEnabled;`

No need to prefix with `_`

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2043,6 +2069,19 @@ public static Object objectForKeyPathComponents(Object[] pathComponents) {
     return messages;
   }
 
+  /**
+   * Set this to true if you want details about the variable assignments
```
lan17: These comments don't seem to correspond to method which is a getter method and not a setter method for a flag?

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2043,6 +2069,19 @@ public static Object objectForKeyPathComponents(Object[] pathComponents) {
...
+   * on the server.
+   * Default is NO.
+   */
+  public static Map<String, Object> variantDebugInfo() {
```
lan17: Create and use custom type for `VariantDebugInfo`

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -664,6 +671,10 @@ private static void startHelper(
     // Get the current inbox messages on the device.
     params.put(Constants.Params.INBOX_MESSAGES, LeanplumInbox.getInstance().messagesIds());
 
+    if (LeanplumInternal.getIsVariantDebugInfoEnabled()) {
+      params.put(Constants.Params.INCLUDE_VARIANT_DEBUG_INFO, true);
```
lan17: You can rewrite this as
`params.put(Constants.Params.INCLUDE_VARIANT_DEBUG_INFO, LeanplumInternal.getIsVariantDebugInfoEnabled());`

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -1924,6 +1941,9 @@ public static void forceContentUpdate(final VariablesChangedCallback callback) {
       Map<String, Object> params = new HashMap<>();
       params.put(Constants.Params.INCLUDE_DEFAULTS, Boolean.toString(false));
       params.put(Constants.Params.INBOX_MESSAGES, LeanplumInbox.getInstance().messagesIds());
+      if (LeanplumInternal.getIsVariantDebugInfoEnabled()) {
```
lan17: You can rewrite this as
params.put(Constants.Params.INCLUDE_VARIANT_DEBUG_INFO, LeanplumInternal.getIsVariantDebugInfoEnabled());

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
e7mac: I do not think we need to create a class for this. This is a special one-off feature and only to be used for debugging. Creating its own class and taking on the serializing from JSON response as well as serializing / deserializing for persistence is extra work that is not needed for this feature

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
e7mac: I used the code style that was present in the file. I think we can do a separate code style pass for the mobile code, but do not want to make this line different than the rest of the file.

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
e7mac: Same as above

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
e7mac: ^

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumInternal.java
e7mac: ^

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
e7mac: ^

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
lan17: It's a good idea to create new classes for data structures such as this as it makes SDK much more user-friendly to use and brings up the quality of code.  Instead of having yourself or customer use an opaque Map class, provide them with self documenting data structure, even if its for debugging.

It should not be much extra work, but would increase code quality by a bit.

It should be done across the board for other things such as `regions`, but can be done in separate review.

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
lan17: I don't see other code in this file using this pattern?

Orthodox Java approach is to use `this` to disambiguate variables when having an arg and member variable share a name.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -1941,6 +1961,12 @@ public void response(JSONObject response) {
               if (response.optBoolean(Constants.Keys.LOGGING_ENABLED, false)) {
                 Constants.loggingEnabled = true;
               }
+
+              Map<String, Object> variantDebugInfo = JsonConverter.mapFromJsonOrDefault(
```
lan17: It seems that this is duplicated from `startHelper`.  Any good way to avoid duplicating same code?

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
lan17: There is a space before `=`.  That is definitely not our style or standard Java code style.

AndroidSDKCore/src/main/java/com/leanplum/internal/VarCache.java
e7mac: Just tried it - can't use "this" cuz we're using static variables. I wanna move this into an instance with singleton, but that's out of scope, so we'll just use the underscore for now

Cyril-Engels: How about Leanplum.variantDebugInfo = variantDebugInfo;

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumInternal.java
e7mac: static var :(

e7mac:
Test this please

e7mac:
closing cuz jenkins is in a stuck state on build machine

PR #215, opened 2018-08-14
Feat(Dockerfile): Add Android emulator and nodejs

sayaansaha:
## Proposed Changes

  - Add Android emulator dependencies into Dockerfile
  - Add nodejs into Dockerfile
  - Pre-build android emulator

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,3 +4,24 @@ RUN apt-get update && \
   apt-get install -y --no-install-recommends \
   build-essential \
   && rm -rf /var/lib/apt/lists/*
+
+ENV PATH="$PATH:$ANDROID_HOME/tools:$ANDROID_HOME/platform-tools:$ANDROID_HOME/tools/bin"
+
+RUN echo $PATH
+
+WORKDIR "/leanplum"
+
+CMD bash
+
+RUN rm -f local.properties
+
+RUN sdkmanager --list
+
+RUN sdkmanager --no_https --licenses
+
+RUN sdkmanager --no_https emulator tools platform-tools "platforms;android-26" "system-images;android-26;google_apis;x86" --verbose
+RUN echo "Creating an Emulator"
+RUN echo no | avdmanager create avd -n "x86" --package "system-images;android-26;google_apis;x86" --tag google_apis
+
+# Needed for adb to work
+EXPOSE 5037 5554 5555
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM jangrewe/gitlab-ci-android
 
 RUN apt-get update && \
   apt-get install -y --no-install-recommends \
+  nodejs \
   build-essential \
   && rm -rf /var/lib/apt/lists/*
 
@@ -25,3 +26,4 @@ RUN echo no | avdmanager create avd -n "x86" --package "system-images;android-26
 
 # Needed for adb to work
 EXPOSE 5037 5554 5555
+


Dockerfile
```diff
@@ -2,5 +2,28 @@ FROM jangrewe/gitlab-ci-android
 
 RUN apt-get update && \
   apt-get install -y --no-install-recommends \
+  nodejs \
```
MaitreyaBuddha: can you lock down version?

Dockerfile
```diff
@@ -2,5 +2,28 @@ FROM jangrewe/gitlab-ci-android
...
+
+WORKDIR "/leanplum"
+
+CMD bash
+
+RUN rm -f local.properties
+
+RUN sdkmanager --list
+
+RUN sdkmanager --no_https --licenses
```
MaitreyaBuddha: lines 10-here are debug and should be removed

Dockerfile
```diff
@@ -2,5 +2,28 @@ FROM jangrewe/gitlab-ci-android
...
+
+RUN sdkmanager --no_https emulator tools platform-tools "platforms;android-26" "system-images;android-26;google_apis;x86" --verbose
+RUN echo "Creating an Emulator"
```
MaitreyaBuddha: No need to run logs in final docker

Dockerfile
```diff
@@ -2,5 +2,28 @@ FROM jangrewe/gitlab-ci-android
...
+RUN echo no | avdmanager create avd -n "x86" --package "system-images;android-26;google_apis;x86" --tag google_apis
```
MaitreyaBuddha: This command and the sdkmanager command both depend on which package and architecture you use. With this in mind, it would behoove us to:

1. Extract names of things like "system-images;android-26;google_apis;x86" to variables.
2. Do the sdkmanager and avdmanager commands in the same `RUN` (like you see in `apt-get`) so they are bundled in the same docker layer.

sayaansaha:
test

Makefile
```diff
@@ -15,6 +15,8 @@ clean-local-properties:
 	rm -f local.properties
 
 sdk: clean-local-properties
+	emulator -verbose -avd device1 -no-skin -no-audio -no-window&
+	sleep 30
```
MaitreyaBuddha: 1. I don't think this is where you want this. This is `make sdk` which should only make the sdk. You do not need an emulator for that.

2. Sleep is a red flag. What are we waiting for? 30 is a magic number. Why 30? If we _have_ to use this we need to explain why. I would like to explore other options first.

Dockerfile
```diff
@@ -1,6 +1,19 @@
...
+
 RUN apt-get update && \
   apt-get install -y --no-install-recommends \
+  nodejs=6.11.4~dfsg-1ubuntu1 \
   build-essential \
   && rm -rf /var/lib/apt/lists/*
+
+ENV PATH="$PATH:$ANDROID_HOME/tools:$ANDROID_HOME/platform-tools:$ANDROID_HOME/tools/bin"
+
+RUN sdkmanager --no_https --licenses && \
```
MaitreyaBuddha: This is not necessary.

Dockerfile
```diff
@@ -1,6 +1,19 @@
...
+  sdkmanager --no_https emulator tools platform-tools \
```
MaitreyaBuddha: As discussed, we cannot use `no_https`

Dockerfile
```diff
@@ -1,6 +1,19 @@
...
+  "platforms;${emulator_platform}" ${system_image} --verbose && \
```
MaitreyaBuddha: You extracted the variable `platform_image` but did not use it

Dockerfile
```diff
@@ -1,6 +1,19 @@
 FROM jangrewe/gitlab-ci-android
 
+ARG emulator_platform=android-26
+ARG system_image="system-images;${emulator_platform};google_apis;x86"
+ARG platform_image="platforms;${emulator_platform}"
```
MaitreyaBuddha: This looks great! Very easy to read and easy to change platform.

Makefile
```diff
@@ -15,6 +16,7 @@ clean-local-properties:
 	rm -f local.properties
 
 sdk: clean-local-properties
+	${EMULATOR} -no-accel -verbose -avd device1 -no-skin -no-audio -no-window&
```
MaitreyaBuddha: This keeps appearing in this PR. This is not necessary to make the `sdk` rule.

sayaansaha: I guess the best place to put it is in a new rule; something like `make e2e tests` rule...

sayaansaha: I removing this from this PR since it is extraneous and opening a new PR for make rules for e2e testing

PR #232, opened 2018-08-28
Correct placement of CI files and other changes that were missed.

MaitreyaBuddha:
The old one was missing some things I had done in experimental branches. This is the correct setup.


Tools/jenkins/build.dockerfile
```diff
@@ -7,13 +7,10 @@ ARG platform_image="platforms;${emulator_platform}"
 RUN apt-get update && \
   apt-get install -y --no-install-recommends \
   nodejs=6.11.4~dfsg-1ubuntu1 \
-  npm \
```
sayaansaha: we needed this because the above nodejs package doesnt for some reason have npm last i tested it. And npm is what we use to install yarn, which is what is needed by our scripts.

sayaansaha:
Also, we need to merge this PR so that i can rebase my release script PR and move the appropriate files into the Tools dir

PR #233, opened 2018-08-29
Use deterministic gradle version in docker, not checked in

MaitreyaBuddha:
Update a few things and remove a bunch of noise. Move gradle installation to docker image


Tools/jenkins/build.dockerfile
```diff
@@ -1,16 +1,36 @@
...
   apt-get install -y --no-install-recommends \
   nodejs=6.11.4~dfsg-1ubuntu1 \
   build-essential \
   && rm -rf /var/lib/apt/lists/*
 
+################################ Gradle Installation ################################
+# Allow the host to use gradle cache, otherwise gradle will always download plugins & artifacts.
+VOLUME ["/root/.gradle/caches/"]
+
+ARG GRADLE_VERSION=4.10
```
sayaansaha: We've had trouble with gradle version 4.9 and 4.10 in the past executing some of the DSL we have in our gradle files. Generally we use Gradle 4.6

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     }
     dependencies {
         classpath 'org.jfrog.buildinfo:build-info-extractor-gradle:4.4.12'
-        classpath 'com.android.tools.build:gradle:3.0.0'
+        classpath 'com.android.tools.build:gradle:3.1.4'
```
sayaansaha: This might or might not cause other peoples Android Studios to complain and ask them to update their tools versions and/or Android Studio to complain. Not that its a big deal, but if we are aware of it we can preempt it.

gradlew
```diff
@@ -1,172 +0,0 @@
-#!/usr/bin/env sh
-
-##############################################################################
-##
-##  Gradle start up script for UN*X
-##
```
sayaansaha: pretty sure this is autogenerated by the gradle tool. I think when you run the gradle wrapper command which android studio does behind the scenes somewhere this script is pasted into the directory.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +0,0 @@
-#Tue Nov 07 13:26:56 CET 2017
-distributionBase=GRADLE_USER_HOME
```
sayaansaha: I believe this file is also autogenerated as well; we might need to add this file to gitignore otherwise it will be accidentally checked in by a dev working on the SDK. Same for file below.

common-methods.gradle
```diff
@@ -1,7 +1,7 @@
 // Export methods by turning them into closures
 ext {
     COMPILE_SDK_VERSION = 27
-    BUILD_TOOLS_VERSION = '27.0.2'
+    BUILD_TOOLS_VERSION = '27.0.3'
```
sayaansaha: Why did we need to change these?

build.sh
```diff
@@ -1,84 +0,0 @@
-#!/usr/bin/env bash
-#
-# LPM | Author: Ben Marten
-# Copyright (c) 2017 Leanplum Inc. All rights reserved.
-#
-set -eo pipefail; [[ $DEBUG ]] && set -x
```
sayaansaha: This essentially means we just deprecated and broke lpm from being able to do an Android release. Which is fine as long as we are aware that this is a breaking change for lpm

Tools/jenkins/build.dockerfile
MaitreyaBuddha: That sounds like a problem with the build system of the path not being deterministic. Our current endeavor is to make a deterministic build system. This passes build and testing in CI. If something is broken, it should be visible in build or test.

build.gradle
MaitreyaBuddha: This is because Android Studio was complaining.

build.sh
MaitreyaBuddha: Yes. Notice this is on `master` and what we are doing is _replacing lpm_. On develop lpm continues to work (or not) as well as it used to.

common-methods.gradle
MaitreyaBuddha: Android studio complained.

gradle/wrapper/gradle-wrapper.properties
MaitreyaBuddha: See line 181 in .gitignore in this review (top file).

gradlew
MaitreyaBuddha: Auto generated by android studio.

Tools/jenkins/build.dockerfile
sayaansaha: Well actually, i meant there are  gradle incompatibilities when executing other commands in the deploy process when using versions higher than 4.9. For example: https://github.com/gradle/gradle/issues/6396, which means we would need to make sure to get the correct corresponding of the other plugin to prevent it. Thats why i was saying using 4.6 is a safe bet because we know that there are no incompatibilities with that particular version with the tools we use. However, this is not really a big issue, it just means that we need to vet all other tools we are using in conjunction with gradle to ensure that they are interoperable.

build.sh
sayaansaha: 👍

gradlew
sayaansaha: 👍

gradle/wrapper/gradle-wrapper.properties
sayaansaha: 👍

PR #234, opened 2018-08-29
Use version text file

MaitreyaBuddha:


AndroidSDKCore/src/main/java/com/leanplum/internal/Constants.java
```diff
@@ -36,7 +36,7 @@
   public static int SOCKET_PORT = 80;
   public static int NETWORK_TIMEOUT_SECONDS = 10;
   public static int NETWORK_TIMEOUT_SECONDS_FOR_DOWNLOADS = 10;
-  public static String LEANPLUM_VERSION = "4.1.1";
+  public static String LEANPLUM_VERSION = BuildConfig.SDK_VERSION + '.' + BuildConfig.BUILD_NUMBER;
```
sayaansaha: Lets not concat these two things together, for example when we have a public release we dont want there to be any build number. Right now you have this defaulted to 0.

Tools/create-release.py
```diff
@@ -1,23 +1,18 @@
...
 import sys 
 
-ANDROID_STRINGS_XML = "AndroidSDK/res/values/strings.xml"
+SDK_VERSION_FILE = "AndroidSDKCore/sdk-version.txt"
 
-def get_current_version(root):
-  for element in root.iter("string"):
-    if element.attrib.get("name") == "sdk_version":
-      return element.text
+def get_current_version():
```
sayaansaha: This super simple, we probably could have explored this kinda solution from the onset.

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
```diff
@@ -1123,6 +1123,29 @@ public void testSyncResources() {
     assertTrue((Boolean) TestClassUtil.getField(FileManager.class, "isInitialized"));
   }
 
+  /**
+   * Test that a valid version is checked in.
```
sayaansaha: Glad you added these tests, but i'm wondering if this file is in fact the best place for these tests.

AndroidSDKCore/src/main/java/com/leanplum/internal/Constants.java
MaitreyaBuddha: I think we decided we _did_ want an _exact_ version number, which includes build. @e7mac was that right? This may have some downstream implications though if places assume "x.x.x" format. Do you know of any specific places, or places to check?

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
MaitreyaBuddha: You caught me. I'll make a new file.

Tools/create-release.py
MaitreyaBuddha: It's only easy because of `file('sdk-version.txt').text` which neither of us knew about at the time ;-)

PR #236, opened 2018-08-30
remove dependency from socketio client

e7mac:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/internal/SocketIOClient.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/internal/SocketIOClient.java
@@ -23,24 +23,23 @@
 
 package com.leanplum.internal;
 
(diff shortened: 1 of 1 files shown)


AndroidSDKCore/src/main/java/com/leanplum/internal/SocketIOClient.java
```diff
@@ -78,15 +77,17 @@ private static String userAgentString() {
         + Constants.LEANPLUM_VERSION + "/" + Constants.LEANPLUM_PACKAGE_IDENTIFIER + ")";
   }
 
-  private static String downloadUriAsString(final HttpUriRequest req)
+  private static String downloadUriAsString(final String urlString)
```
MaitreyaBuddha: Can we write a mock and test this somehow?

PR #240, opened 2018-09-05
Fix release script to have everything we need

MaitreyaBuddha:
Release script needed to add changes and push correctly.

### DIFF
--- a/Tools/create-release.bash
+++ b/Tools/create-release.bash
@@ -18,5 +18,5 @@ RELEASE_BRANCH=release/${release_version}
 git checkout -b ${RELEASE_BRANCH}
 
 git add AndroidSDKCore/sdk-version.txt
-git commit -m "Created release ${release_version}"
+git commit -m "Created ${TYPE} release ${release_version}"
 git push --set-upstream origin ${RELEASE_BRANCH}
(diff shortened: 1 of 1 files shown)


Tools/create-release.bash
```diff
@@ -11,7 +11,12 @@ set -o errexit
 TYPE=$1
 
 release_version=$(./Tools/create-release.py ${TYPE})
+cat AndroidSDKCore/sdk-version.txt
```
e7mac: what would this line do?

e7mac: Are we adding the version in this file somehow?

Tools/create-release.bash
```diff
@@ -11,7 +11,12 @@ set -o errexit
...
 
-# create a branch, push on success
-git checkout -b "release/${release_version}"
-git push --set-upstream origin "release/${release_version}"
+git reset HEAD --
+RELEASE_BRANCH=release/${release_version}
+git checkout -b ${RELEASE_BRANCH}
+
+git add AndroidSDKCore/sdk-version.txt
+git commit -m "Created ${TYPE} release ${release_version}"
```
Cyril-Engels: `git commit -m "Create ${TYPE} release ${release_version}."` Present tense?

Tools/create-release.bash
MaitreyaBuddha: Yes that file is already in to note our current version. It's read during build time to be hard-coded in the binary. This line `cat...` Just displays the version. It's a debug line but I think useful.

PR #242, opened 2018-09-11
LP-9741 make heartbeat methods idempotent

e7mac:
currently, every time `startHeartbeat` is called, we create a new executor that fires a method every 15 minutes. We want this to only happen once

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
@@ -1042,16 +1042,7 @@ public class Leanplum {
   private static void startHeartbeat() {
     synchronized (heartbeatLock) {
       if (heartbeatExecutor == null) {
-        heartbeatExecutor = Executors.newSingleThreadScheduledExecutor();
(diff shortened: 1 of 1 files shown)


AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -1041,23 +1041,26 @@ private static void resumeInternal() {
...
     synchronized (heartbeatLock) {
-      heartbeatExecutor = Executors.newSingleThreadScheduledExecutor();
-      heartbeatExecutor.scheduleAtFixedRate(new Runnable() {
-        public void run() {
-          try {
-            Request.post(Constants.Methods.HEARTBEAT, null).sendIfDelayed();
-          } catch (Throwable t) {
-            Util.handleException(t);
+      if (heartbeatExecutor == null) {
+        heartbeatExecutor = Executors.newSingleThreadScheduledExecutor();
```
MaitreyaBuddha: This is getting a little tabbed out to unreadability. Can we extract a `createHeartbeat` function to call `if (heartbeatExecutor == null)` ?

PR #245, opened 2018-09-19
feature flagging on client

gg4race:


AndroidSDKCore/src/main/java/com/leanplum/internal/FeatureFlagManager.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public void setEnabledFeatureFlags(HashSet<String> enabledFeatureFlags) {
+        this.enabledFeatureFlags= enabledFeatureFlags;
+    }
+
+    public Boolean isFeatureFlagEnabled(String featureFlagName) {
+        return this.enabledFeatureFlags.contains(featureFlagName);
+    }
+
+    public HashSet<String> getEnabledFeatureFlags() {
```
e7mac: let's get rid of this method. We don't need a public method for this

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2172,4 +2173,13 @@ private static void parseSdkCounters(JSONObject response) {
       CountAggregator.INSTANCE.setEnabledCounters(counterSet);
     }
   }
+
+  private static void parseFeatureFlags(JSONObject response) {
```
e7mac: let's add a test for this method and a similar one for countAggregator

AndroidSDKCore/src/main/java/com/leanplum/internal/FeatureFlagManager.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.HashSet;
+
+public class FeatureFlagManager {
+    public static final FeatureFlagManager INSTANCE = new FeatureFlagManager();
+
+    private HashSet<String> enabledFeatureFlags = new HashSet<>();
+
+    public void setEnabledFeatureFlags(HashSet<String> enabledFeatureFlags) {
+        this.enabledFeatureFlags= enabledFeatureFlags;
```
MaitreyaBuddha: nit: spaces around operators.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2165,11 +2168,38 @@ public static void clearUserContent() {
...
+      countAggregator.setEnabledCounters(counterSet);
+    }
+  }
+
+  private static void parseFeatureFlags(JSONObject response) {
+    parseFeatureFlags(response, FeatureFlagManager.INSTANCE);
+  }
+
+  @VisibleForTesting
+  public static void parseFeatureFlags(JSONObject response, FeatureFlagManager featureFlagManager) {
```
e7mac: currently this method is parsing feature flags AND setting them on the featureFlagManager. Instead, lets have this method parse the feature flags and return a List<String>. Then, we can test this method easily, and set that list to the enabled near line 864.

gg4race: counters and flags need to be set above the line 864 block or else shouldParse... tests would fail, I set them in lines 824 and 826 instead, similar to iOS structure

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -818,7 +820,10 @@ protected Void doInBackground(Void... params) {
               Constants.loggingEnabled = true;
             }
 
-            parseSdkCounters(response);
+            HashSet<String> enabledCounters = parseSdkCounters(response);
```
e7mac: tip:
When defining types, use `Set<x>` instead of the concrete class `HashSet<x>`. Even though it doesn't matter a lot in this particular instance, you build the habit of retaining as much flexibility as possible. By defining it as `Set` you retain the flexibility of how exactly to implement the set. However, if you define it as a `HashSet`, then if you find a more optimized solution in the future, you can't simply change the backing implementation

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2164,12 +2169,33 @@ public static void clearUserContent() {
     VarCache.clearUserContent();
   }
 
-  private static void parseSdkCounters(JSONObject response) {
+  @VisibleForTesting
+  public static HashSet<String> parseSdkCounters(JSONObject response) {
     JSONArray enabledCounters = response.optJSONArray(
             Constants.Keys.ENABLED_COUNTERS);
+    HashSet<String> counterSet = new HashSet<>();
```
e7mac: so going by the previous logic, this line should be rewritten as:
`Set<String> counterSet = new HashSet<>();`
so we're still using `HashSet` in the implementation, but if needed it could change to a different `Set` type

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2164,12 +2169,33 @@ public static void clearUserContent() {
...
+            Constants.Keys.ENABLED_FEATURE_FLAGS);
+    HashSet<String> featureFlagSet = new HashSet<>();
+    if (enabledFeatureFlags != null) {
+      featureFlagSet.addAll(toList(enabledFeatureFlags));
+    }
+    return featureFlagSet;
+  }
+
+  private static List<String> toList(JSONArray array) {
+      List<String> list = new ArrayList<>();
```
e7mac: here, you are doing it correctly for the `List` type

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
```diff
@@ -1267,4 +1273,62 @@ public void testGetDeviceId() {
...
+   * Tests for parsing feature flags
+   */
+  @Test
+  public void testStartResponseShouldParseFeatureFlags() {
+    // Setup sdk.
+    setupSDK(mContext, "/responses/simple_start_response.json");
+
+    assertEquals(true, FeatureFlagManager.INSTANCE.isFeatureFlagEnabled("testFeatureFlag1"));
+    assertEquals(true, FeatureFlagManager.INSTANCE.isFeatureFlagEnabled("testFeatureFlag2"));
+  }
```
e7mac: let's add a line - this ensures that we aren't returning true for everything
`assertEquals(false, FeatureFlagManager.INSTANCE.isFeatureFlagEnabled("missingFeatureFlag"));`

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2164,12 +2169,33 @@ public static void clearUserContent() {
...
+    JSONArray enabledFeatureFlags = response.optJSONArray(
+            Constants.Keys.ENABLED_FEATURE_FLAGS);
+    HashSet<String> featureFlagSet = new HashSet<>();
+    if (enabledFeatureFlags != null) {
+      featureFlagSet.addAll(toList(enabledFeatureFlags));
+    }
+    return featureFlagSet;
+  }
+
+  private static List<String> toList(JSONArray array) {
```
toanvc: Hey, why don't you just use 
```
private static HashSet<String> toHashSet(JSONArray array) {
      HashSet<String> yourset = new HashSet<>();
      for (int i = 0; i < array.length(); i++) {
          yourset.add(array.optString(i));
      }
      return yourset;
}
```
The way converting from `JSONArray` => `List` => add all to `HashSet` is taking time and you can do it more easier.

AndroidSDKCore/src/main/java/com/leanplum/internal/FeatureFlagManager.java
```diff
@@ -0,0 +1,19 @@
+package com.leanplum.internal;
+
+import android.support.annotation.VisibleForTesting;
+
+import java.util.HashSet;
+
+public class FeatureFlagManager {
+    public static final FeatureFlagManager INSTANCE = new FeatureFlagManager();
```
toanvc: If you use static INSTANCE, should you make the constructor `private`? It looks like you're using `Singleton`.

gg4race: Thanks for the review Toan! 👍

gg4race:
i looked at the logs, not sure it has to do with any changes I made?

e7mac:
test this please

PR #248, opened 2018-09-25
Create new UUID when sending partial batch

e7mac:
The UUIDs set to each request identifies which "batch" it is a part of. However, when we send smaller batches (due to OOM), we want to identify the sub-batch as a different batch to the servers.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
@@ -707,8 +707,17 @@ public class Request {
       editor.remove(Constants.Defaults.UUID_KEY);
       SharedPreferencesUtil.commitChanges(editor);
     }
-
+    if (fraction < 1) {
+      setNewUUIDs(requestData);
+    }
     return requestData;
+  }
+
+  private void setNewUUIDs(List<Map<String, Object>> requests) {
+    String uuid = UUID.randomUUID().toString();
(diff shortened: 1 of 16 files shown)


AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -707,10 +707,19 @@ static void deleteSentRequests(int requestsCount) {
...
       SharedPreferencesUtil.commitChanges(editor);
     }
-
+    if (fraction < 1) {
+      setNewUUIDs(requestData);
+    }
     return requestData;
   }
 
+  private void setNewUUIDs(List<Map<String, Object>> requests) {
```
MaitreyaBuddha: This is actually more like `setBatchUUID`. There is nothing inherently "new" about this function, and there is only one UUID. You set a batch UUID in another location for the original batch too, right? Should both use the same code to make this DRY?

e7mac: When requests are added to the database, they are assigned a UUID, say `uuid-1`. This is also stored in sharedPreferences and re-used until the batch is sent.

However, when this function is called, we are overwriting the old UUID to a "new" one, say `uuid-2`. The reason we are doing this is because this is called in the case where we only sent a fraction of the requests from the batch that had `uuid-1`. 

We can't really call this from the first line, because at that point it's not setting it to a given batch of requests, but one by one.

PR #252, opened 2018-10-01
Fix requests to send

e7mac:
As per https://leanplum.atlassian.net/browse/LP-9792, Kinedu is running into the situation where they are getting more responses than the requests it is expecting to send. 

Irrelevant start calls are removed and we expect the responses to be equal to the number of "relevant" requests. However, the JSON encoding that is done includes those calls. 

So we are getting more responses than the expected requests, since we are not pruning the JSON to only includes requests to send.

In the future, the pruning of these irrelevant start calls should be done in a better way


e7mac:
ah.. it was still a WIP :)

PR #255, opened 2018-10-03
add request id as parameter to every request

e7mac:


AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -723,11 +724,27 @@ private static void handleApiResponse(JSONObject response, List<Map<String, Obje
       for (int i = requests.size() - 1; i >= 0; i--) {
         Map<String, Object> currentRequest = requests.get(i);
         if (Constants.Methods.START.equals(currentRequest.get(Constants.Params.ACTION))) {
-          if (i < responseCount) {
-            lastStartResponse = Request.getResponseAt(response, i);
+          if (currentRequest.containsKey(Constants.Params.REQ_ID)) {
+              for (int j = Request.numResponses(response) - 1; j >= 0; j--) {
```
aafrooze: out of curiosity, why are you looping from `n-1` to `0` instead of the other way?

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -723,11 +724,27 @@ private static void handleApiResponse(JSONObject response, List<Map<String, Obje
...
+                          currentRequest.get(Constants.Params.REQ_ID)) {
+                    lastStartResponse = currentResponse;
+                    hasStartResponse = true;
+                    break;
+                  }
+                } catch (JSONException e) {
+                  Util.handleException(e);
+              }
+            }
+          } else {
```
aafrooze: I think we should be handling backwards compatibility on the server, in case we introduce any bugs it'll be easier to fix. Thoughts?

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestFactory.java
```diff
@@ -21,20 +21,55 @@
...
   public Request createRequest(
       String httpMethod, String apiMethod, Map<String, Object> params) {
-    return new Request(httpMethod, apiMethod, params);
+    return new Request(httpMethod, apiMethod, params, incrementedRequestId());
+  }
+
+  private int incrementedRequestId() {
+    requestId = (requestId + 1) % MAX_REQUEST_ID;
+    saveRequestIdToDisk(requestId);
+    return requestId;
```
aafrooze: do we need the `requestId` to be an integer? As far as we're concerned this just needs to be unique _within_ the same batch that's being submitted to the server and a uuid gives that to us.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -723,11 +724,27 @@ private static void handleApiResponse(JSONObject response, List<Map<String, Obje
       for (int i = requests.size() - 1; i >= 0; i--) {
         Map<String, Object> currentRequest = requests.get(i);
         if (Constants.Methods.START.equals(currentRequest.get(Constants.Params.ACTION))) {
-          if (i < responseCount) {
-            lastStartResponse = Request.getResponseAt(response, i);
+          if (currentRequest.containsKey(Constants.Params.REQ_ID)) {
+              for (int j = Request.numResponses(response) - 1; j >= 0; j--) {
+                JSONObject currentResponse = Request.getResponseAt(response, i);
```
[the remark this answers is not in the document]
e7mac: good catch

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
e7mac: server will just pass through whatever `reqId` it gets, so that shouldnt be a problem either way... since the client will also use the older logic if the request id's dont match

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestFactory.java
e7mac: uuid is the easiest for sure. Kelly is concerned about the costs and wants this to be the smallest possible. This is a balance between uniqueness and smallest payload

aafrooze: I'm glad we're considering cost in our implementation and design, but we're talking about a few dollars here and it's not worth sacrificing reliability (using a standard for generating global ids) and maintainability (maintaining the added code, although it's pretty straight forward ) for.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
aafrooze: 👍 so this code will never get executed right?

e7mac: yup, exactly.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
e7mac: gracefully degrade - If we end up having multiple requests with the same ID, we pick the latest request with that id.

AndroidSDKCore/src/main/java/com/leanplum/internal/Constants.java
```diff
@@ -157,6 +157,7 @@ public static boolean isNoop() {
     public static final String INBOX_MESSAGE_ID = "newsfeedMessageId";
     public static final String INBOX_MESSAGES = "newsfeedMessages";
     public static final String PARAMS = "params";
+    public static final String REQ_ID= "reqId";
```
MaitreyaBuddha: I don't think it's "consistency" to keep putting constants here. I think you can put it in a better scoped place.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -723,6 +724,21 @@ private static void handleApiResponse(JSONObject response, List<Map<String, Obje
       for (int i = requests.size() - 1; i >= 0; i--) {
         Map<String, Object> currentRequest = requests.get(i);
         if (Constants.Methods.START.equals(currentRequest.get(Constants.Params.ACTION))) {
+          if (currentRequest.containsKey(Constants.Params.REQ_ID)) {
+              for (int j = Request.numResponses(response) - 1; j >= 0; j--) {
+                JSONObject currentResponse = Request.getResponseAt(response, j);
+                try {
+                  if (currentResponse.get(Constants.Params.REQ_ID) ==
+                          currentRequest.get(Constants.Params.REQ_ID)) {
+                    lastStartResponse = currentResponse;
```
MaitreyaBuddha: Should we do the check on 742 first? It's just going to overwrite `lastStartResponse` when `i < responseCount`. I can't tell off-hand how often/likely that is. Also we could skip this loop.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
MaitreyaBuddha: Could this loop be 10k long? Do we have any speculation on performance?

AndroidSDKCore/src/main/java/com/leanplum/internal/Constants.java
e7mac: Can't think of a better place for this particular one. Ideas?

MaitreyaBuddha: Request.java seems perfect to me. Also, we are using `"reqId"` for efficiency in our requests. This does not mean we want to use abbreviated variable names, which is against LP standards. We can call this `REQUEST_ID_KEY`. Even though its content is `"reqId"`, the name is still fully descriptive.

PR #262, opened 2018-10-09
change directory for getting jar file

e7mac:
gradle plugin 3.1.4 puts the intermediate jar in a different location than the gradle plugin 3.0.2

Since we've upgraded the gradle version previously, this update fixes the makeJar task


e7mac:
gradle plugin 3.1.4 puts the intermediate jar in a different location than the gradle plugin 3.0.2

Since we've upgraded the gradle version previously, this update fixes the `makeJar` task

PR #270, opened 2018-10-17
callback on message displayed

e7mac:
https://leanplum.atlassian.net/browse/E2-1658

In order to let the client know an in-app message is shown, we need to expose a callback for developers to hook into.


common-methods.gradle
```diff
@@ -11,7 +11,7 @@ ext {
     LEANPLUM_FCM_ARTIFACT_ID = 'leanplum-fcm'
     LEANPLUM_GCM_ARTIFACT_ID = 'leanplum-gcm'
     LEANPLUM_LOCATION_ARTIFACT_ID = 'leanplum-location'
-    LEANPLUM_SDK_VERSION = "$System.env.ANDROID_VERSION_STRING"
+    LEANPLUM_SDK_VERSION = "4.2.0"
```
MaitreyaBuddha: This looks incorrect.

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
```diff
@@ -1455,4 +1457,21 @@ public void testParseLastStartResponseGivenMultipleStartShouldReturnResponse() t
...
+
+    MessageDisplayedCallback callback = Mockito.mock(MessageDisplayedCallback.class);
+
+    Leanplum.addMessageDisplayedHandler(callback);
+    Leanplum.triggerMessageDisplayed(testActionContext);
+
+    Mockito.verify(callback).setActionContext(testActionContext);
+    Mockito.verify(callback).run();
+  }
+
```
MaitreyaBuddha: nit: extra line.

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
```diff
@@ -1455,4 +1457,21 @@ public void testParseLastStartResponseGivenMultipleStartShouldReturnResponse() t
     }
     return responsesList;
   }
+
+  /**
+   * Test trigger message displayed calls callback
+   */
+  @Test
+  public void testTriggerMessageDisplayed() {
```
MaitreyaBuddha: This tests a callback, not display, right? Misnomer? Also can we test that the callback has all the information from the business requirements for archiving?

AndroidSDKCore/src/main/java/com/leanplum/ActionContext.java
```diff
@@ -390,6 +390,7 @@ private boolean createActionContextForMessageId(String messageAction,
           public void variablesChanged() {
             try {
               ActionManager.getInstance().recordMessageImpression(actionContext.getMessageId());
+              Leanplum.triggerMessageDisplayed(actionContext);
```
MaitreyaBuddha: All three of these happen at the same place as all calls to `ActionManager.getInstance().recordMessageImpression`, so it seems they are inexorable tied together. I think it would make more sense to:
1. Put this inside `ActionManager.getInstance().recordMessageImpression`
2. Make a function that contains both the new callback and `ActionManager.getInstance().recordMessageImpression`.

I don't like the idea of keeping two functions in sync manually.

AndroidSDKCore/src/main/java/com/leanplum/callbacks/MessageDisplayedCallback.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public void setActionContext(ActionContext actionContext) {
+    this.actionContext = actionContext;
+  }
+
+  public void run() {
+    this.messageDisplayed(actionContext);
+  }
+
+  public abstract void messageDisplayed(ActionContext actionContext);
```
MaitreyaBuddha: Does `ActionContext` mean anything to the developer consuming this function? Since this is specifically for archiving messages, maybe we want something that makes more sense, like a `MessageContext` with members for each business requirement of the archival feature?

e7mac: That's something I was thinking about too. 
Either 
1. this callback ends up being specifically for archiving messages OR
2. this is a more general callback with more information than just archiving messages

Since it's a customer facing callback, if more information is needed later, making changes and deprecating becomes a bit messy.

I was thinking we enable the most info possible to be accessible through the callback and  document the things specifically needed for the archiving use case. That way if we have other use cases or if someone needs more information for archiving, it all continues to funnel through the same callback.

Thoughts?

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
e7mac: To keep the method name short, I put the specific information in the comment above. `triggerMessageDisplayed` is the specific method being tested here. I can make the method name more descriptive again.

AndroidSDKCore/src/main/java/com/leanplum/ActionContext.java
e7mac: Agreed.
I thought it made more sense to keep the `triggerMessageDisplayed` and move the `recordMessageImpression` under that. The `triggerMessageDisplayed` is now responsible for recording impression and calling any user callbacks.

AndroidSDKCore/src/main/java/com/leanplum/callbacks/MessageDisplayedCallback.java
MaitreyaBuddha: I like the idea of having all information given in a generic, more scalable way. However, in reality we have only one exact use-case. We know exactly what is needed for a specific purpose. I don't want to have a paragraph of documentation explaining how to use a generic API that was only created for this specific use-case. Less setup, more plug-and-play, more delighting our customer.

At the risk of scope-creep... Could we add a function to ActionContext like `getBlob/Archive/FINRA/etc...` that returns what is needed for this exact use-case?

Otherwise I would lean toward option 1. If we ever need something more generic we won't have to deprecate 1. We can just add the more generic one.

AndroidSDKCore/src/main/java/com/leanplum/callbacks/MessageDisplayedCallback.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public void setDeliveryDateTime(Date deliveryDateTime) {
+    this.deliveryDateTime = deliveryDateTime;
+  }
+
+  public void run() {
+    this.messageDisplayed(messageID, messageBody, recipientUserID, deliveryDateTime);
+  }
+
+  public abstract void messageDisplayed(String messageID, String messageBody,
```
MaitreyaBuddha: I don't want to stop this up with scope creep again, but if we really think about our customer here this could be more useful. They want a blob of this information that they will store somewhere. If we had a POJO representing the "message" with all the required fields, it would be scalable for the future and give them an easy way to deal with it/JSON-ify it. Now they have to hard-code creating their own object into this function signature that would require a code change from them to update.

Also FYI, it's always a red flag to have many arguments in a function signature.

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
```diff
@@ -1455,4 +1460,33 @@ public void testParseLastStartResponseGivenMultipleStartShouldReturnResponse() t
...
+      public void messageDisplayed(String messageID, String messageBody, String recipientUserID, Date deliveryDateTime) {
+        assertTrue(messageID.equals(testMessageID));
+        assertTrue(messageBody.equals(testMessageBody));
+        assertTrue(recipientUserID.equals(testUserID));
+        long timeDiff = new Date().getTime() - deliveryDateTime.getTime();
+        assertTrue(timeDiff < 1000);
+      }
+    };
+    Leanplum.addMessageDisplayedHandler(callback);
+    Leanplum.triggerMessageDisplayed(testActionContext);
```
MaitreyaBuddha: If this breaks and does not actually call the callback, this test will pass.

PR #275, opened 2018-10-24
Final tools for creating and deploying artifacts

MaitreyaBuddha:


build.gradle
```diff
@@ -2,8 +2,8 @@
 buildscript {
     repositories {
         mavenCentral()
-        jcenter()
         google()
+        jcenter()
```
MaitreyaBuddha: jcenter reliability is vulgar. Not sure if this fixed prioritizing it (it should have been lower than maven already) or if it just stopped being broken.

PR #278, opened 2018-11-06
Use name instead of message for counter name

MaitreyaBuddha:
Per spec https://leanplum.atlassian.net/wiki/spaces/ENG/pages/624394247/SDK+Counting

The arg is "name" for counter name, not "message".


gg4race:
I think we are creating params for LOG object or LOG method here, in which the name of the counter is the message of the log. I think that makes sense?

PR #285, opened 2018-12-13
E2-10024 multi-channel geofence

gg4race:

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
@@ -25,7 +25,6 @@ import android.app.Activity;
 import android.content.Context;
 import android.location.Location;
 import android.os.AsyncTask;
-import android.os.Message;
 import android.support.annotation.VisibleForTesting;
 import android.text.TextUtils;
 
@@ -54,7 +53,7 @@ import com.leanplum.internal.Util;
 import com.leanplum.internal.Util.DeviceIdInfo;
 import com.leanplum.internal.VarCache;
 import com.leanplum.messagetemplates.MessageTemplates;
-import com.leanplum.models.GeofenceEventName;
+import com.leanplum.models.GeofenceEventType;
 import com.leanplum.models.MessageArchiveData;
 import com.leanplum.utils.BuildUtil;
 import com.leanplum.utils.SharedPreferencesUtil;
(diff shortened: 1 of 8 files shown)


AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumInternal.java
```diff
@@ -270,36 +270,55 @@ public void variablesChanged() {
     }
   }
 
+  private static Map<String, Object> makeTrackArgs(final String event, double value, String info,
```
hrishileanplum: 👍  great for code reusability.

AndroidSDKLocation/src/main/java/com/leanplum/LocationManagerImplementation.java
```diff
@@ -306,14 +307,16 @@ private void updateTrackedGeofences() {
...
               // triggering it, as a locally triggered push notification.
               && !backgroundGeofences.contains(geofence)) {
-            Number lastStatus = (Number) stateBeforeBackground.get(geofence.getRequestId());
-            Number currentStatus = (Number) lastKnownState.get(geofence.getRequestId());
+            Number lastStatus = (Number) stateBeforeBackground.get(geofenceId);
+            Number currentStatus = (Number) lastKnownState.get(geofenceId);
             if (currentStatus != null && lastStatus != null) {
               if (GeofenceStatus.shouldTriggerEnteredGeofence(lastStatus, currentStatus)) {
                 maybePerformActions(geofence, "enterRegion");
+                Leanplum.trackGeofence("enter_region", geofenceId);
```
hrishileanplum: "enter_region" will it help if this is enum? Would like to know full context.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -1858,7 +1859,7 @@ public static void track(String event, double value, String info) {
    * strings or numbers. You can use up to 200 different parameter names in your app.
    */
 
-  public static void trackGeofence(String event, String info) {
+  public static void trackGeofence(GeofenceEventName event, String info) {
```
hrishileanplum: Thank you @gg4race for the changes. Can you please let me know whats the use of "info" here? Is this any string that the customers can input or this can also be an enum?

hrishileanplum: Not a biggy. Also can you please rename this to "GeofenceEventType", instead of "GeofenceEventTypeName". Sounds like a different type of geofence events.

gg4race: the info would be the name of the geofence that is crossed. These are custom names defined by the customer so probably don't make sense to be enum

hrishileanplum: sounds great.

PR #288, opened 2019-01-02
E2-1736: Message archive handles all kinds of text

e7mac:
For Push message archiving, I had assumed that message["body"] is always a string. However, this is an incorrect assumption. For certain message types, message["body"] is a dictionary. We need to change the code to handle all the different kinds of messages.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
@@ -1362,16 +1362,16 @@ public class Leanplum {
   @VisibleForTesting
   public static String messageBodyFromContext(ActionContext actionContext) {
     Object messageObject =  actionContext.getArgs().get("Message");
-    if (messageObject.getClass().isInstance(String.class)) {
+    if (messageObject instanceof String) {
       return (String) messageObject;
     } else {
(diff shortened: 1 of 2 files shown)


AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -1337,24 +1338,46 @@ public static void triggerMessageDisplayed(ActionContext actionContext) {
...
+    Object messageObject =  actionContext.getArgs().get("Message");
+    if (messageObject instanceof String) {
+      return (String) messageObject;
+    } else {
+      HashMap<String, String> messageDict = (HashMap<String, String>) messageObject;
+      if (messageDict.get("Text") != null &&
+              messageDict.get("Text") instanceof String) {
+        return messageDict.get("Text");
+      }
+      if (messageDict.get("Text value") != null &&
```
hrishileanplum: I am curious how this Dictionary key was implemented with a space and caps "Text value" , we should definitely try to standardize this.

e7mac: this is the current key we're getting from the server. Agree on standardizing

PR #302, opened 2019-01-23
remove socket dependency

e7mac:
ApacheHttp library was deprecated in Api 28. We want to remove the dependency on this library and move to other recommended android ways to make url connections


AndroidSDKCore/src/main/java/com/leanplum/internal/SocketIOClient.java
```diff
@@ -78,15 +77,17 @@ private static String userAgentString() {
         + Constants.LEANPLUM_VERSION + "/" + Constants.LEANPLUM_PACKAGE_IDENTIFIER + ")";
   }
 
-  private static String downloadUriAsString(final HttpUriRequest req)
+  private static String downloadUriAsString(final String urlString)
```
hrishileanplum: Looks like the change is to private method. But, i do not see the callers of these method change to the new signature. Can we please make sure that callers are updated and the code runs all test cases successfully?

AndroidSDKCore/src/main/java/com/leanplum/internal/SocketIOClient.java
```diff
@@ -236,9 +237,8 @@ public void connect() {
       return;
     new Thread() {
       public void run() {
-        HttpPost post = new HttpPost(mURL);
         try {
-          String line = downloadUriAsString(post);
+          String line = downloadUriAsString(mURL);
```
hrishileanplum: Is this the only location for the call?

AndroidSDKCore/src/main/java/com/leanplum/internal/SocketIOClient.java
```diff
@@ -78,15 +77,17 @@ private static String userAgentString() {
...
-  private static String downloadUriAsString(final HttpUriRequest req)
+  private static String downloadUriAsString(final String urlString)
       throws IOException {
-    AndroidHttpClient client = AndroidHttpClient.newInstance(userAgentString());
-    try {
-      HttpResponse res = client.execute(req);
-      return readToEnd(res.getEntity().getContent());
-    } finally {
-      client.close();
+    String fullString = "";
```
hrishileanplum: Just a suggestion, but your approach is fine as well with "+" operator since it internally uses StringBuilder.
Instead of String fullString = ""; 

You could also use StringBuilder.

StringBuilder result = new StringBuilder();
String line;
while((line = reader.readLine()) != null) {
    result.append(line);
}
System.out.println(result.toString());

AndroidSDKCore/src/main/java/com/leanplum/internal/SocketIOClient.java
e7mac: this is the only location, yes

PR #304, opened 2019-01-23
update gradle major version

e7mac:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/messagetemplates/BaseMessageDialog.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/messagetemplates/BaseMessageDialog.java
@@ -55,6 +55,7 @@ import android.widget.TextView;
 
 import com.leanplum.ActionContext;
 import com.leanplum.Leanplum;
+import com.leanplum.core.R;
 import com.leanplum.utils.BitmapUtil;
 import com.leanplum.utils.SizeUtil;
 import com.leanplum.views.BackgroundImageView;
@@ -105,7 +106,7 @@ public class BaseMessageDialog extends Dialog {
     dialogView.setLayoutParams(layoutParams);
 
     RelativeLayout view = createContainerView(activity, fullscreen);
-    view.setId(108);
+    view.setId(R.id.container_view);
     dialogView.addView(view, view.getLayoutParams());
 
     if ((!isWeb || (webOptions != null && webOptions.hasDismissButton())) && !isHtml) {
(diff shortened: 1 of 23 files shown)


AndroidSDKCore/src/main/java/com/leanplum/messagetemplates/BaseMessageDialog.java
```diff
@@ -105,7 +106,7 @@ protected BaseMessageDialog(Activity activity, boolean fullscreen, BaseMessageOp
     dialogView.setLayoutParams(layoutParams);
 
     RelativeLayout view = createContainerView(activity, fullscreen);
-    view.setId(108);
```
hrishileanplum: Wow

AndroidSDKCore/src/main/java/com/leanplum/messagetemplates/BaseMessageDialog.java
```diff
@@ -209,7 +210,7 @@ public void run() {
 
   private CloseButton createCloseButton(Activity context, boolean fullscreen, View parent) {
     CloseButton closeButton = new CloseButton(context);
-    closeButton.setId(103);
```
hrishileanplum: Thank you for these changes :)

AndroidSDKPush/src/main/java/com/leanplum/LeanplumNotificationHelper.java
```diff
@@ -299,6 +300,7 @@ static void setModifiedBigPictureStyle(Notification.Builder notificationBuilder,
    * @param messageText String with text for push notification.
    * @return Notification.BigPictureStyle or null.
    */
+  @TargetApi(16)
```
hrishileanplum: Should these be changed from 16-26 as well?

AndroidSDKPush/src/main/java/com/leanplum/LeanplumNotificationHelper.java
```diff
@@ -299,6 +300,7 @@ static void setModifiedBigPictureStyle(Notification.Builder notificationBuilder,
...
   static Notification.BigPictureStyle getBigPictureStyle(Bundle message, Bitmap bigPicture,
       String title, final String messageText) {
     if (Build.VERSION.SDK_INT < 16 || message == null || bigPicture == null) {
```
hrishileanplum: Yea, i see mix of api target 16 or 26. Any ideas why?

AndroidSDKPush/src/main/java/com/leanplum/LeanplumNotificationHelper.java
e7mac: Yeah. I'm putting the lowest needed TargetApi numbers in here. 
Here's why
1. The gradle compilation process checks to ensure that we aren't using any android calls that are below the minimum supported version (14 in this case)
2. However, in certain methods, if we are handling the different API calls ourselves (through branching etc), then we can let the compiler know that and ask it for a higher API check.
3. In this case, I'm setting the TargetApi wrt the branching logic in the function, 16 in most cases, but 26 for notification channels.

PR #318, opened 2019-01-30
Lp 10190 support gzip

hrishileanplum:
Fixes #
Adding gzip encoding support. This will help in compressed data response between sdk and the backend.

## Proposed Changes
 Adding gzip request support in the code and handling the gzip response differently based on the backend response header contract.


AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
```diff
@@ -557,6 +557,10 @@ static HttpURLConnection createHttpUrlConnection(
         getApplicationName(context) + "/" + getVersionName() + "/" + RequestOld.appId() + "/" +
             Constants.CLIENT + "/" + Constants.LEANPLUM_VERSION + "/" + getSystemName() + "/" +
             getSystemVersion() + "/" + Constants.LEANPLUM_PACKAGE_IDENTIFIER);
+    if (Leanplum.featureFlagManager().isFeatureFlagEnabled(Leanplum.featureFlagManager().FEATURE_FLAG_REQUEST_GZIP)) {
+      urlConnection.setRequestProperty("Accept-Encoding", "gzip");
+      Leanplum.countAggregator().incrementCount("operation_with_request_gzip");
+    }
```
e7mac:
```suggestion
    } else {
      Leanplum.countAggregator().incrementCount("operation_without_request_gzip");
    }
```

will let us monitor the rollout

AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
```diff
@@ -556,7 +557,8 @@ static HttpURLConnection createHttpUrlConnection(
     urlConnection.setRequestProperty("User-Agent",
         getApplicationName(context) + "/" + getVersionName() + "/" + RequestOld.appId() + "/" +
             Constants.CLIENT + "/" + Constants.LEANPLUM_VERSION + "/" + getSystemName() + "/" +
-            getSystemVersion() + "/" + Constants.LEANPLUM_PACKAGE_IDENTIFIER);
+            getSystemVersion() + "/" + Constants.LEANPLUM_SUPPORTED_ENCODING + "/" + Constants.LEANPLUM_PACKAGE_IDENTIFIER);
+    urlConnection.setRequestProperty("Accept-Encoding", Constants.LEANPLUM_SUPPORTED_ENCODING);
```
aafrooze: Could you add a comment here as to why we're appending "gzip" to the end of the user agent?
This is done for compatibility with AppEngine Standard:
1. Must include `Accept-Encoding: gzip` in the heaer
2. Must include the phrase `gzip` in the `User-Agent` header
https://cloud.google.com/appengine/kb/

AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
```diff
@@ -660,7 +662,12 @@ public static void saveResponse(URLConnection op, OutputStream outputStream) thr
   private static String getResponse(HttpURLConnection op) throws IOException {
     InputStream inputStream;
     if (op.getResponseCode() < 400) {
-      inputStream = op.getInputStream();
+      String contentHeader = op.getHeaderField("content-encoding");
```
aafrooze: Does `getHeaderField` trim white spaces too?

AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
```diff
@@ -660,7 +662,12 @@ public static void saveResponse(URLConnection op, OutputStream outputStream) thr
...
+      if (contentHeader != null && contentHeader.equalsIgnoreCase(Constants.LEANPLUM_SUPPORTED_ENCODING)) {
+        inputStream = new GZIPInputStream(op.getInputStream());
```
aafrooze: What happens here if the server returns plain text but with the `content-encoding: gzip` response header?
What happens if the content is compressed, but the header is missing?

AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
hrishileanplum: sure

AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
hrishileanplum: 1. What happens here if the server returns plain text but with the `content-encoding: gzip` response header?
A: Code will crash.As we discussed yesterday, the server has to respect the contract.

2. What happens if the content is compressed, but the header is missing?
A: Code will crash.As we discussed yesterday, the server has to respect the contract.

AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
hrishileanplum: "content-encoding" is the hardcoded key, we are looking for. Do you mean if the return is "g zip" or " gzip " as value? Does it handle that?
Can you confirm, is that what you mean, and do we need to check that?

PR #332, opened 2019-02-08
Gradle upgrade and Travis cleanup

e7mac:


Makefile
```diff
@@ -4,23 +4,6 @@
 #
 ####################################################################
 
-testSDK:
```
hrishileanplum: I feel that we might need local deployment mechanism as a backup in case travis fails. We cannot completely relay on travis to test, build and deploy. 
So, looks like this PR has gradle upgrade and removing of make commands only correct? Can we keep the Make commands as it is?

e7mac: I'd like to remove the make commands cuz they are an unnecessary layer. I'd prefer us (and new people) to have the knowledge of doing `./gradlew test` or `./gradlew assembleRelease`depending on what we want to do.

PR #334, opened 2019-02-09
Add tag verification during deploy

e7mac:
When deploying, we want the build machine to fail if the tag and the sdk-version.txt files are out of sync


Makefile
```diff
@@ -4,7 +4,10 @@
...
 ####################################################################
 
+verifyTag:
+	./Tools/verifyTag.sh
+
 tagCommit:
 	git tag `cat sdk-version.txt`; git push origin `cat sdk-version.txt`
 
-deploy: tagCommit
+deploy: verifyTag tagCommit
```
hrishileanplum: This is good safety net. Little confused with name. Can we rename "deploy" to "verifyTagAndCommit" ?

e7mac: The name is coming from the fact that this is what we are doing for deployment. If you look at the iOS makefile, it also updates the constants file. We can make another task called `verifyTagAndCommit` but I want the deployment to be a single make deploy

PR #337, opened 2019-02-20
Adding cache update if variants are changed

milos1290:
Fixes #

## Proposed Changes

  -
  -
  -


AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -981,6 +981,7 @@ private static void applyContentInResponse(JSONObject response, boolean alwaysAp
         || !messages.equals(VarCache.getMessageDiffs())
         || !updateRules.equals(VarCache.getUpdateRuleDiffs())
         || !eventRules.equals(VarCache.getEventRuleDiffs())
+        || !variants.equals(VarCache.variants())
```
hrishileanplum: Can you give me quick context why we need this variants? Also, any reasons, we dont have test cases for this?

PR #338, opened 2019-02-25
E2-1807 feature flag for crash monitoring

e7mac:


AndroidSDKCore/src/main/java/com/leanplum/internal/FeatureFlagManager.java
```diff
@@ -10,8 +10,6 @@
 public class FeatureFlagManager {
```
hrishileanplum: Now, that we are accessing FeatureFlags directly , do we still need this class? FeatureFlagManager?

AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
```diff
@@ -849,7 +849,10 @@ public static void initExceptionHandling(Context context) {
    * Handles uncaught exceptions in the SDK.
    */
   public static void handleException(Throwable t) {
-    ExceptionHandler.getInstance().reportException(t);
+    if (FeatureFlagManager.INSTANCE.isFeatureFlagEnabled(
```
hrishileanplum: Can we add test cases for this PR covering the feature flag please? And this check please

AndroidSDKCore/src/main/java/com/leanplum/internal/FeatureFlagManager.java
e7mac: FeatureFlagManager is the code that actually parses the response and keeps track of which flags are enabled. The FeatureFlag class is just where we keep track of the actual feature flag strings.
Changing where the constant strings are stored does not obviate the logic that parses the backend response and performs logic to check against it.
Is refactoring the strings to the different file somehow communicating that we no longer need the logic? Is there a better way that you have in mind that we can implement?

AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
e7mac: The code that I have added basically take the previously existing line `ExceptionHandler.getInstance().reportException(t);` and wrap it in a code block that additionally calls `FeatureFlagManager.INSTANCE.isFeatureFlagEnabled()`. Can you help me understand which line you want tested? 

I can see one of 2 options:
1. Test the previously existing line `ExceptionHandler.getInstance().reportException(t)`
2. Test the newly added line - FeatureFlagManager.INSTANCE.isFeatureFlagEnabled(). If so, this method is part of the `FeatureFlagManager` class and hence is tested in `FeatureFlagManagerTests.java`
3. Test whether the method FeatureFlagManager.INSTANCE.isFeatureFlagEnabled() works for the specific argument that we are calling here `FeatureFlag.FEATURE_FLAG_CRASH_MONITORING`. This would be a case of writing a test for the heck of it and I'd like to decide as a team, if that's the direction we want to go in

AndroidSDKCore/src/main/java/com/leanplum/internal/FeatureFlagManager.java
hrishileanplum: 👍

AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
hrishileanplum: I would say 1 simple test would be to just check, if the method gets called, when the feature flag is enabled. Just check if mock feature flag enabled, is the method getting called.
(FeatureFlagManager.INSTANCE.isFeatureFlagEnabled(
            FeatureFlag.FEATURE_FLAG_CRASH_MONITORING))

PR #340, opened 2019-02-26
set new batch UUID when partial events are sent

e7mac:
https://leanplum.atlassian.net/browse/LP-10291

Design doc here: https://docs.google.com/document/d/1BRy6MXS1qKeHKzYGjox0J4CtCBLqVHZ3lfet3M17OtQ/edit#


AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
```diff
@@ -734,10 +734,21 @@ static void deleteSentRequests(int requestsCount) {
       editor.remove(Constants.Defaults.UUID_KEY);
       SharedPreferencesUtil.commitChanges(editor);
     }
-
+    // if we send less than 100% of requests, we need to reset the batch
```
hrishileanplum: Can you please help me understand this flow. So, if the fraction is less than 100%, we will generate new BatchID, but that would be sent in a next api call? Or when will the new batch of calls be made? When the app gets killed and relaunched etc?

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
```diff
@@ -734,10 +734,21 @@ static void deleteSentRequests(int requestsCount) {
...
+    if (fraction < 1) {
+      setNewBatchUUID(requestData);
+    }
     return requestData;
   }
 
+  private void setNewBatchUUID(List<Map<String, Object>> requests) {
+    String uuid = UUID.randomUUID().toString();
+    for (Map<String, Object> request : requests) {
+      request.put(UUID_KEY, uuid);
```
hrishileanplum: Dont we need to set the new uui_key in SharedPrefs? Also, when we put in "Request", can you please help me understand, does it mean we will make that many individual calls?
I thought we had batching before making api calls.

AndroidSDKTests/src/test/java/com/leanplum/internal/RequestOldUtilTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        // Mock this so async things run synchronously
+        ReflectionHelpers.setStaticField(Util.class, "singleThreadExecutor", new SynchronousExecutor());
+    }
+
+    @Test
+    public void testSetNewBatchUUID() {
+        LeanplumEventDataManager.init(Leanplum.getContext());
+
+        RequestOld request1 = new RequestOld("POST", Constants.Methods.START, null);
```
aafrooze: `new RequestOld(RequestOldUtilTest.POST ...` or get rid of the constant?

AndroidSDKTests/src/test/java/com/leanplum/internal/RequestOldUtilTest.java
```diff
@@ -0,0 +1,64 @@
...
+        request1.sendEventually();
+        RequestOld request2 = new RequestOld("POST", Constants.Methods.TRACK, null);
```
aafrooze: `new RequestOld(RequestOldUtilTest.POST ...` or get rid of the constant?

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
e7mac: At the beginning of this code block, all the requests will have the same UUID. When the fraction is 100%, everything gets sent with that UUID and the UUID gets reset. However, if we send a batch thats smaller, the same UUID can no longer be re-used and a new one must be generated for the next send.
This is all persisted so app launch is irrelevant.

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
e7mac: Lemme check if we need to set the new key in SharedPrefs. 
Putting something in Request doesn't mean it will be an individual call. It is all batched still (which is why this UUID must be set in the first place)

e7mac: @hrishileanplum I saved the new UUID in shared prefs as well now. Do pls lmk if its clear now

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOldUtil.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class RequestOldUtil {
+
+    public void setNewBatchUUID(List<Map<String, Object>> requests) {
+        String uuid = UUID.randomUUID().toString();
+        Context context = Leanplum.getContext();
+        SharedPreferences preferences = context.getSharedPreferences(
+                LEANPLUM, Context.MODE_PRIVATE);
+        SharedPreferences.Editor editor = preferences.edit();
+        editor.putString(Constants.Defaults.UUID_KEY, uuid);
```
hrishileanplum: SharedPrefs for storing BatchId, but it it used for some reason? Do we have a get batchId from sharedprefs? Wanted to understand the use case for sharedprefs.
Also, if we are doing getSharedPrefs for some reason, not sure when. Will it be a problem, is say the API fall fails?

e7mac: The batchId is gotten from SharedPrefs when new requests are added. https://github.com/Leanplum/Leanplum-Android-SDK/blob/b86e546598f6f28c8f7809567327204484c3dea8/AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java#L158

e7mac: I have not changed anything about API call failures in this PR. I am not sure what the exact mechanism is but you can check it out in `RequestOld.java` 
I know that when we send a batch, is when we change the UUID that's stored in sharedprefs. So that's the model I am following

PR #345, opened 2019-03-19
E2 1802 fcm gcm

hrishileanplum:
Fixes #
The goal of this ticket is to remove GCM support from Leanplum Android sdk. As starting APRIL 11th, GCM will no longer be supported by google.

## Proposed Changes
https://leanplum.atlassian.net/browse/E2-1802

Cleanup the code, get rid of GCM and refactor to use the new correct FCM standard for fetching tokens.


Makefile
```diff
@@ -1,10 +1,10 @@
 ####################################################################
-#
+#
```
e7mac: could we revert all the changes to this file?

settings.gradle
```diff
@@ -2,6 +2,5 @@ include ':AndroidSDKCore'
 include ':AndroidSDKLocation'
 include ':AndroidSDKPush'
```
e7mac: are we thinking of combining leanplum-push and leanplum-fcm in a future PR ?

Makefile
hrishileanplum: sure

hrishileanplum: Actually now i realized. i had got merge conflicts with this file somehow. Thats the reason.

settings.gradle
hrishileanplum: Yes. In future for sure or in our new architecture . I tried doing it. But, there is lot more changes than i expected. So, not for this commit.

milos1290:
Looking good, just one note, since we are modifying public API we need to bump major version.

PR #348, opened 2019-04-01
E2-1790 android socket fix

gg4race:
Fix 1:
Originally we do various checks to see whether our connection is secure, like

`int port = (mURI.getPort() != -1) ? mURI.getPort() : (mURI.getScheme().equals("wss") ? 443 : 80);`

However wss is not supported in our URI java library so getScheme() returns http or https, so the condition is aways false. Before, we were using http, so checking for "wss" should return false, we were being passed to the correct port. However once we switched to https we expect these checks to return true, but since checking "wss" returns false, we are being passed to the wrong path.

The fix is to check for https instead of wss.

Fix 2:
We were not resetting flag `connecting` to false when connection fails, so even though we are supposed to retry connecting every few seconds, we are not currently doing that. This is a fix for that.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/internal/Socket.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/internal/Socket.java
@@ -81,6 +81,9 @@ public class Socket {
       @Override
       public void onError(Exception error) {
         Log.e("Development socket error", error);
+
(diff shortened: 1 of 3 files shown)


e7mac:
looks great! Are there any tests we can add for this?

gg4race:
added some tests, build successful!!

PR #355, opened 2019-04-25
E2-1815: download files directly from URL instead of API call

e7mac:
As per https://leanplum.atlassian.net/browse/E2-1815 (Proposal 1), we will be sending the files key with the URL for each file. If this exists, we want the SDK to use the URL directly to get these files.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
@@ -811,8 +811,8 @@ public class Leanplum {
             if (variants == null) {
               Log.d("No variants received from the server.");
             }
-            Map<String, String> filenamesToURL = parseFileURLs(response);
-            FileManager.setFilenameToURL(filenamesToURL);
+            Map<String, String> filenameToURLs = parseFilenameToURLs(response);
+            FileManager.setFilenameToURLs(filenameToURLs);
 
             if (BuildUtil.isNotificationChannelSupported(context)) {
               // Get notification channels and groups.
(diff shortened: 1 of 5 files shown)


AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2303,6 +2307,13 @@ public static void clearUserContent() {
     return featureFlagSet;
   }
 
+  @VisibleForTesting
+  public static Map<String, String> parseFileURLs(JSONObject response) {
+    JSONObject filesObject = response.optJSONObject(
```
hrishileanplum: I think you need try catch here? Also null check, if the json key does not exists?

AndroidSDKCore/src/main/java/com/leanplum/internal/FileManager.java
```diff
@@ -185,6 +189,10 @@ public static boolean fileExistsAtPath(String path) {
     return path != null && new File(path).exists();
   }
 
+  public static void setFilenameToURL(Map<String, String> filenameToURL) {
+    FileManager.filenameToURL = filenameToURL;
```
hrishileanplum: Just similar to previous iOS note on naming. Lets name it filenamesToUrls

AndroidSDKCore/src/main/java/com/leanplum/internal/FileManager.java
```diff
@@ -108,6 +109,9 @@ public static DownloadFileResult maybeDownloadFile(boolean isResource, String st
       if (!FileManager.fileExistsAtPath(realPath)) {
         realPath = FileManager.fileRelativeToDocuments(stringValue);
         if (!FileManager.fileExistsAtPath(realPath)) {
+          if (FileManager.filenameToURL != null && FileManager.filenameToURL.containsKey(stringValue) && urlValue == null) {
+            urlValue = FileManager.filenameToURL.get(stringValue);
```
e7mac: I tried adding a test here, but the structure of code made it very difficult

AndroidSDKCore/src/main/java/com/leanplum/internal/FileManager.java
hrishileanplum: setFilenameToURL -? setFilenameToURLs

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
e7mac: good "catch" :), thanks!

PR #356, opened 2019-04-30
Lp 10446 start call back variable call back

hrishileanplum:
Fixes #
Fix for https://leanplum.atlassian.net/browse/LP-10446
The offline mode was detected using Util.isConnected(), but callbacks were not getting called.

## Proposed Changes
  -On offline mode, the change is propagated to call the callback handlers with false as success value.
  -
  -

### DIFF
--- a/AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
+++ b/AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
@@ -1671,4 +1671,63 @@ public class LeanplumTest extends AbstractTest {
 
     assertEquals(4, VarCache.variants().size());
   }
+
+  @Test
+  public void testStartAndVariableChangeCallBacksForOffline() throws Exception {
+    final Semaphore semaphore = new Semaphore(1);
+    semaphore.acquire();
+
+    //Offline Mode.
+     ResponseHelper.seedResponseNull("/responses/simple_start_response.json");
+
+    // Expected request params.
+    final HashMap<String, Object> expectedRequestParams = CollectionUtil.newHashMap(
+        "city", "(detect)",
+        "country", "(detect)",
+        "location", "(detect)",
+        "region", "(detect)",
+        "locale", "en_US"
+    );
+
+    // Validate request.
+    // Validate request.
+    RequestHelper.addRequestHandler(new RequestHelper.RequestHandler() {
+      @Override
+      public void onRequest(String httpMethod, String apiMethod, Map<String, Object> params) {
+        assertEquals(Constants.Methods.START, apiMethod);
+        assertTrue(params.keySet().containsAll(expectedRequestParams.keySet()));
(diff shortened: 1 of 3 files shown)


AndroidSDKTests/src/test/java/com/leanplum/_whitebox/utilities/ResponseHelper.java
```diff
@@ -74,4 +74,19 @@ public static void seedResponse(String filename) {
...
     }
   }
+
+  /**
+   * Seeds the Null response to Util.getResponse method.
+   *
+   */
+
+
+  public static void seedResponseNull(String filename) {
```
aafrooze: don't need a file here since you're only setting `null` as the response

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
```diff
@@ -1671,4 +1671,63 @@ public void variablesChanged() {
...
+      public void onResponse(boolean success) {
+        assertTrue(success);
+        semaphore.release();
+      }
+    });
+    assertTrue(Leanplum.hasStarted());
+
+    semaphore.acquire();
+
+    // Seed getVars response.
```
e7mac: please could we split up these 2 responses into their own test files? It'd be great to keep the tests testing as little individually as possible so that when something breaks, we don't have to parse the test to understand what happened

AndroidSDKTests/src/test/java/com/leanplum/_whitebox/utilities/ResponseHelper.java
```diff
@@ -74,4 +74,19 @@ public static void seedResponse(String filename) {
...
+  /**
+   * Seeds the Null response to Util.getResponse method.
+   *
+   */
+
+
+  public static void seedResponseNull() {
+    try {
+      doReturn(false).when(Util.class, Util.isConnected());
+      //doReturn(parseResponse(filename)).when(Util.class, "getResponse", anyObject());
```
e7mac: please could we just remove this completely, instead of commenting out?

AndroidSDKTests/src/test/java/com/leanplum/_whitebox/utilities/ResponseHelper.java
hrishileanplum: done

AndroidSDKTests/src/test/java/com/leanplum/LeanplumTest.java
hrishileanplum: done

AndroidSDKTests/src/test/java/com/leanplum/_whitebox/utilities/ResponseHelper.java
hrishileanplum: Nice catch. done. my bad actually.

PR #360, opened 2019-05-07
E2-1893: add counters to methods we want to deprecate

e7mac:


AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2246,6 +2247,7 @@ public static void setDeviceLocation(Location location, LeanplumLocationAccuracy
           "call setDeviceLocation. If you prefer to always set location manually, " +
           "then call disableLocationCollection.");
     }
+    Leanplum.countAggregator().incrementCount("setDeviceLocation,type");
```
gg4race: well, if all the other counters use colons, maybe we should here too?

PR #366, opened 2019-05-29
E2-1930 Fixing a crash loop when receiving push notification

milos1290:
What              | Where/Who
JIRA Issue        | [E2-1930](https://leanplum.atlassian.net/browse/E2-1930)

## Background
Crash when receiving push notification.
## Implementation
Making sure context is not null when receiving push notification.

### DIFF
--- a/AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
+++ b/AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
@@ -61,6 +61,9 @@ public class LeanplumPushFirebaseMessagingService extends FirebaseMessagingServi
   @Override
   public void onMessageReceived(RemoteMessage remoteMessage) {
     try {
+      if (Leanplum.getContext() == null) {
+        Leanplum.setApplicationContext(this);
+      }
       Map<String, String> messageMap = remoteMessage.getData();
       if (messageMap.containsKey(Constants.Keys.PUSH_MESSAGE_TEXT)) {
         LeanplumPushService.handleNotification(this, getBundle(messageMap));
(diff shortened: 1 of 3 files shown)


AndroidSDKTests/src/test/java/com/leanplum/internal/RequestOldTest.java
```diff
@@ -371,6 +372,34 @@ public void testGetRequestsWithEncodedStringStoredRequests() {
...
+
+    final Semaphore semaphore = new Semaphore(1);
+    semaphore.tryAcquire();
+
+    // Given a request.
+    Map<String, Object> params = new HashMap<>();
+    params.put("data1", "value1");
+    params.put("data2", "value2");
+    RequestOld request = new RequestOld(POST, Constants.Methods.START, params);
+    request.onError(new RequestOld.ErrorCallback() {
```
aafrooze: Will this get called in this instance? If we're returning early in `saveRequestForLater`, then you won't be throwing an error, right?

milos1290: it will throws error first then tries to save for later

AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
```diff
@@ -61,6 +61,9 @@ public void onNewToken(String token) {
   @Override
   public void onMessageReceived(RemoteMessage remoteMessage) {
     try {
+      if (Leanplum.getContext() == null) {
```
hrishileanplum: Do we know why Leanplum.getContext() is null?  Otherwise with limited context on context this change looks good to me.

milos1290: Yes, we set the context when `start` call is issued, in case push notification is received and we are not started we can get in a situation where context is not set yet and/or `start` call not called

PR #367, opened 2019-06-19
E2-1956: add de-dupe logic

e7mac:
What              | Where/Who
JIRA Issue        | [E2-1957](https://leanplum.atlassian.net/browse/E2-1957)
People Involved   | @shivanshukumar , @hrishileanplum , @aafrooze 

## Background
https://leanplum.atlassian.net/wiki/spaces/E2/pages/806944897/Design+doc+for+de-duping+push+notifications+on+the+device+Android+iOS

## Implementation
adding a dictionary to de-dupe with a 1 hour TTL

## Testing steps
Manual

Yes

### DIFF
--- a/AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
+++ b/AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
@@ -22,15 +22,20 @@
 package com.leanplum;
 
 import android.annotation.SuppressLint;
+import android.content.Context;
+import android.content.SharedPreferences;
 import android.os.Build;
 import android.os.Bundle;
 
 import com.google.firebase.messaging.FirebaseMessagingService;
 import com.google.firebase.messaging.RemoteMessage;
 import com.leanplum.internal.Constants;
+import com.leanplum.internal.JsonConverter;
 import com.leanplum.internal.Log;
 import com.leanplum.internal.Util;
 
+import java.util.Date;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -65,6 +70,9 @@ public class LeanplumPushFirebaseMessagingService extends FirebaseMessagingServi
         Leanplum.setApplicationContext(this);
       }
       Map<String, String> messageMap = remoteMessage.getData();
+      if (isDuplicateNotification(messageMap)) {
+        return;
+      }
       if (messageMap.containsKey(Constants.Keys.PUSH_MESSAGE_TEXT)) {
         LeanplumPushService.handleNotification(this, getBundle(messageMap));
       }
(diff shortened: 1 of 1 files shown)


AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
```diff
@@ -22,15 +22,21 @@
 package com.leanplum;
 
 import android.annotation.SuppressLint;
```
hrishileanplum: As per the doc, for iOS there was existing duplicate notification logic. Were you able to investigate the background why it was implemented. The file was LPActionManager.m

AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
```diff
@@ -65,6 +71,13 @@ public void onMessageReceived(RemoteMessage remoteMessage) {
         Leanplum.setApplicationContext(this);
       }
       Map<String, String> messageMap = remoteMessage.getData();
+      try {
+        if (isDuplicateNotification(messageMap)) {
```
hrishileanplum: We have existing method LPActionManager in LPActionManager.m, which needs to be addressed as well,  otherwise we will have 2 duplicate clearing methods. Which will be wrong.

Also, i would not recommend checking for duplicates on "onMessageRecieved", rather implement this logic in LPActionManager, where the existing dedup implementation is there. This does not gurantee that the previous message was displayed.

AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
```diff
@@ -86,4 +99,63 @@ private Bundle getBundle(Map<String, String> messageMap) {
     }
     return bundle;
   }
+
+  private Boolean isDuplicateNotification(Map<String, String> messageMap) {
+    Boolean isDuplicate = false;
+    String messageId = "";
+    if (messageMap.containsKey("_lpn")) {
```
hrishileanplum: This needs checking. Please reference the existing duplicate resolution code. Looks like we might be using the backend provided message Id. Also, will go through this code in person, little confused here.

AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
```diff
@@ -86,4 +99,63 @@ private Bundle getBundle(Map<String, String> messageMap) {
...
+    if (handledNotifications.containsKey(messageId)) {
+      isDuplicate = true;
+    } else {
+      handledNotifications.put(messageId, new Date().getTime());
+    }
+    persistHandledNotifications(handledNotifications);
+    return isDuplicate;
+  }
+
+  private Map<String, Object> retrieveHandledNotifications() {
```
hrishileanplum: The plan as per the doc was using VarCache, because it handles encryption and decryption code. Also, does have the same point of responsibility of storing anything caching.

AndroidSDKFcm/src/main/java/com/leanplum/LeanplumPushFirebaseMessagingService.java
```diff
@@ -86,4 +99,63 @@ private Bundle getBundle(Map<String, String> messageMap) {
...
+    Map<String, Object> handledNotifications;
+    if (jsonString != null) {
+      handledNotifications = JsonConverter.fromJson(jsonString);
+    } else {
+      handledNotifications = new HashMap<>();
+    }
+    return handledNotifications;
+  }
+
+  private void persistHandledNotifications(Map<String, Object> handledNotifications) {
```
hrishileanplum: Definitely VarCache needs to be used. This code should not be in this file of firebase.

PR #368, opened 2019-06-24
Fixing DB not initialized when PN arrives

milos1290:
What              | Where/Who
JIRA Issue        | [JIRA_TICKET_ID](https://leanplum.atlassian.net/browse/E2-1970)
People Involved   | @milos1290, @hrishileanplum 

## Background
This PR fixes 2 issues
1) when PN arrives and we need to pull new content, DB is not initialised because initialisation happens in `start` and app is not running.
2) when PN arrives we are offloading a network request from IntentService to an async task, service will be killed and kill our process.

## Implementation

1) Making DB manager singleton
2) conditionally decide whether to run task async or on current thread.

## Testing steps

* Install the app
* Kill the app from task manager
* Create IAM
* Send push that chains to IAM from previous step

Yes

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
@@ -625,7 +625,7 @@ public class Leanplum {
 
   private static void startHelper(
       String userId, final Map<String, ?> attributes, final boolean isBackground) {
-    LeanplumEventDataManager.init(context);
+    LeanplumEventDataManager.sharedInstance();
     checkAndStartNotificationsModules();
     Boolean limitAdTracking = null;
     String deviceId = RequestOld.deviceId();
--- a/AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
@@ -48,38 +48,43 @@ import java.util.UUID;
  * @author Anna Orlova
  */
 public class LeanplumEventDataManager {
+
+  private static LeanplumEventDataManager instance;
+
   private static final String DATABASE_NAME = "__leanplum.db";
   private static final int DATABASE_VERSION = 1;
   private static final String EVENT_TABLE_NAME = "event";
   private static final String COLUMN_DATA = "data";
   private static final String KEY_ROWID = "rowid";
 
-  private static SQLiteDatabase database;
-  private static LeanplumDataBaseManager databaseManager;
-  private static ContentValues contentValues = new ContentValues();
+  private SQLiteDatabase database;
+  private LeanplumDataBaseManager databaseManager;
+  private ContentValues contentValues = new ContentValues();
+  private boolean sendErrorLogs = false;
 
-  static boolean willSendErrorLog = false;
-
-  /**
-   * Creates connection to database, if database is not present, it will automatically create it.
-   *
-   * @param context Current context.
-   */
-  public static void init(Context context) {
-    if (database != null) {
-      Log.e("Database is already initialized.");
-      return;
-    }
-
-    // Create database if needed.
+  private LeanplumEventDataManager() {
     try {
+      Context context = Leanplum.getContext();
+
+      if (context == null) {
+        return;
+      }
+
       if (databaseManager == null) {
-        databaseManager = new LeanplumDataBaseManager(context);
+        databaseManager = new LeanplumDataBaseManager(Leanplum.getContext());
       }
       database = databaseManager.getWritableDatabase();
+
     } catch (Throwable t) {
       handleSQLiteError("Cannot create database.", t);
     }
+  }
+
+  public static LeanplumEventDataManager sharedInstance() {
+    if (instance == null) {
+      instance = new LeanplumEventDataManager();
+    }
+    return instance;
   }
 
   /**
@@ -87,14 +92,14 @@ public class LeanplumEventDataManager {
    *
    * @param event String with json of event.
    */
-  static void insertEvent(String event) {
+  void insertEvent(String event) {
     if (database == null) {
       return;
     }
     contentValues.put(COLUMN_DATA, event);
     try {
       database.insert(EVENT_TABLE_NAME, null, contentValues);
-      willSendErrorLog = false;
+      sendErrorLogs = false;
     } catch (Throwable t) {
       handleSQLiteError("Unable to insert event to database.", t);
     }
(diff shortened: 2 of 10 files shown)


AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
```diff
@@ -48,53 +48,58 @@
...
   private static final String COLUMN_DATA = "data";
   private static final String KEY_ROWID = "rowid";
 
-  private static SQLiteDatabase database;
-  private static LeanplumDataBaseManager databaseManager;
-  private static ContentValues contentValues = new ContentValues();
+  private SQLiteDatabase database;
+  private LeanplumDataBaseManager databaseManager;
+  private ContentValues contentValues = new ContentValues();
+  private boolean sendErrorLogs = false;
```
hrishileanplum: Any reason for "sendErrorLogs" rename?

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
```diff
@@ -48,53 +48,58 @@
...
+    }
+    return instance;
+  }
+
   /**
    * Inserts event to event table.
    *
    * @param event String with json of event.
    */
-  static void insertEvent(String event) {
```
hrishileanplum: To confirm here. So since LeanplumEventDataManager is singleton, you dont need static?

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
```diff
@@ -158,33 +163,41 @@ static void deleteEvents(int count) {
...
     } catch (Throwable t) {
       handleSQLiteError("Unable to get a number of rows in the table.", t);
     }
     return count;
   }
 
+  /**
+   * Whether we are going to send error log or not.
+   */
+  boolean willSendErrorLogs() {
```
hrishileanplum: Why is this needed since its default value? Where do we change the value .

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
```diff
@@ -239,6 +252,8 @@ private static void migrateFromSharedPreferences(SQLiteDatabase db) {
 
         editor.remove(Constants.Defaults.COUNT_KEY);
 
+        ContentValues contentValues = new ContentValues();
```
hrishileanplum: Can you please explain the need of ContentValues here?

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
```diff
@@ -516,20 +520,29 @@ private void sendNow() {
...
+          } catch (Throwable t) {
+            Util.handleException(t);
+          }
+          return null;
         }
-        return null;
+      });
+    } else {
+      try {
+        sendRequests();
```
hrishileanplum: Both if and else do the same thing?           try {
            sendRequests();
          } catch (Throwable t) {
            Util.handleException(t);
          }

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -625,7 +625,7 @@ private static void checkAndStartNotificationsModules() {
 
   private static void startHelper(
       String userId, final Map<String, ?> attributes, final boolean isBackground) {
-    LeanplumEventDataManager.init(context);
+    LeanplumEventDataManager.sharedInstance();
```
hrishileanplum: Can you please explain why initialization wasnt happening? I see init here correct

AndroidSDKTests/src/test/java/com/leanplum/internal/RequestOldTest.java
```diff
@@ -262,7 +262,7 @@ public void testRemoveIrrelevantBackgroundStartRequests() throws NoSuchMethodExc
   @Test
   public void testJsonEncodeUnsentRequestsWithExceptionLargeNumbers() throws NoSuchMethodException,
```
hrishileanplum: You need extra test cases, specific to initialization.

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
milos1290: Just to make it consistent, nothing else.

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
milos1290: Yeah, its singleton now and no need for static methods.

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
milos1290: We use the flag in case we fail to save data into SQL so we could send error logs from database

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
milos1290: It used static one from outer class, since its a private variable now its not accessible anymore.

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
milos1290: Not really, the first one is offloading request into background task, the else part is running it on current thread (which is background thread)

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
milos1290: As you can see, initialization happens in `startHelper` which is actually causing the issues when app is not running (i.e push notification received).

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
hrishileanplum: Can you please let me know why is that point above? Why couldnt we just run in background thread, no matter what the parent thread is?

milos1290: Yeah it doesn't really matter, we can always create new async task, but in that case I'll need to synchronise call to forceContentUpdate in LeanplumPushService, to make sure that IntentService doesn't get killed.

hrishileanplum: I would prefer that approach please. Its way more cleaner than handling these If else conditions. Can you please make that change?

PR #378, opened 2019-11-18
[LP-11063] Fixing batching of start call and race condition

milos1290:
What              | Where/Who
JIRA Issue        | [LP-11063](https://leanplum.atlassian.net/browse/LP-11063)]

## Background
Fixes an issue with start call getting batched and not executed immediately.
Fixes an issue where event will be saved and not sent in next batch.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
@@ -486,19 +486,20 @@ public class RequestOld implements Requesting {
       return;
     }
 
-    this.sendEventually();
+    // We need to save request first.
+    sendEventually();
 
     Leanplum.countAggregator().incrementCount("send_now");
 
-    Util.executeAsyncTask(true, new AsyncTask<Void, Void, Void>() {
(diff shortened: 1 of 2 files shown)


AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
```diff
@@ -486,19 +486,20 @@ private void sendNow() {
       return;
     }
 
-    this.sendEventually();
+    // We need to save request first.
+    sendEventually();
```
hrishileanplum: whats this change for?  this.sendEventually(); -> sendEventually(); ?

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
```diff
@@ -486,19 +486,20 @@ private void sendNow() {
...
 
     Leanplum.countAggregator().incrementCount("send_now");
 
-    Util.executeAsyncTask(true, new AsyncTask<Void, Void, Void>() {
```
hrishileanplum: Why are you changing the RequestOld class?

PR #380, opened 2019-12-05
Fixed Start response getting discarded with parallel requests

Nezz:
Fixes `start` response getting discarded when issuing another API call like `getVars` before `start` finishes.

## Proposed Changes

  - Removed logic that discards the `start` response when another API call was executed.
  - Renamed `handleApiResponse` to `handleStartResponse` as the function is only used to handle responses to the `start` call, not any API call.
  - Fixed comment that's supposed to explain what `request.onApiResponse(null)` does.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
@@ -705,7 +705,7 @@ public class Leanplum {
     request.onApiResponse(new RequestOld.ApiResponseCallback() {
       @Override
       public void response(List<Map<String, Object>> requests, JSONObject response, int countOfEvents) {
-        Leanplum.handleApiResponse(response, requests, request, countOfEvents);
+        Leanplum.handleStartResponse(response, requests, request, countOfEvents);
       }
     });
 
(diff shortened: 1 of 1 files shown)


AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -718,27 +718,20 @@ public void response(List<Map<String, Object>> requests, JSONObject response, in
     LeanplumInternal.triggerStartIssued();
   }
 
-  private static void handleApiResponse(JSONObject response, List<Map<String, Object>> requests,
```
hrishileanplum: Is there a Jira ticket associated with this  PR?

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -718,27 +718,20 @@ public void response(List<Map<String, Object>> requests, JSONObject response, in
...
+  private static void handleStartResponse(JSONObject response, List<Map<String, Object>> requests,
                                         final RequestOld request, int countOfUnsentRequests) {
     JSONObject lastStartResponse = null;
 
-    // Find and handle the last start response.
     try {
-      // Checks if START event inside the current batch. If database index of START event bigger
-      // then a number of count of events that we got from the database - decrease START event
```
hrishileanplum: Can you please explain what is happening here and how deleting this code helps?

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
Nezz: We have a Zendesk ticket where we reported the issue in November. See ticket #29522.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
Nezz: The issue can be reproduced when another request is made (e.g. `getVars`) while `start` didn't receive a response yet.
In such cases the request’s `dataBaseIndex` is 1 and `countOfUnsentRequests` is also 1:
image001
This causes `return` to get called and the response of `start` is discarded. This will cause the callback for `start` never to get fired, causing the app to wait indefinitely for the callback to get fired.

The Zendesk ticket contains a more details description of the issue.

milos1290:
@Nezz  I'm closing this PR since it was fixed by refactoring how response callbacks were tracked and executed ( #383 )

PR #385, opened 2020-02-05
Using OperationQueue

milos1290:
What              | Where/Who
JIRA Issue        | [LP-9360](https://leanplum.atlassian.net/browse/LP-9360)

## Background
Removes the OsHandler class in favor of OperationQueue. Removes usage of AsyncTasks in favor of executors. Improves test execution speed


AndroidSDKTests/src/test/java/com/leanplum/internal/ShadowOperationQueue.java
```diff
@@ -16,4 +16,9 @@ public boolean addOperation(Runnable operation) {
     public void addParallelOperation(Runnable operation) {
         operation.run();
     }
+
+    @Override
+    public void addUiOperation(Runnable operation) {
```
e7mac: does this need to be running on the main thread? Is that something that we can make happen?

milos1290: yeah, all public callbacks are executed on main thread

PR #390, opened 2020-03-31
[LP-11240] Adding support for androidx Advertising Client

milos1290:
What              | Where/Who
JIRA Issue        | [LP-11240](https://leanplum.atlassian.net/browse/LP-11240)

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
@@ -36,7 +36,6 @@ import android.net.wifi.WifiInfo;
 import android.net.wifi.WifiManager;
 import android.os.Build;
 import android.provider.Settings.Secure;
-import androidx.annotation.RequiresPermission;
 import android.text.TextUtils;
 import android.util.TypedValue;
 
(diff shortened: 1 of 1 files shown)


AndroidSDKCore/src/main/java/com/leanplum/internal/Util.java
```diff
@@ -194,23 +197,40 @@ private static String getWifiMacAddressHash(Context context) {
...
+
+      for (String name : classNames) {
+        try {
+          Object adInfo = Class.forName(name)
+              .getMethod("getAdvertisingIdInfo", Context.class)
+              .invoke(null, caller);
+
+          if (name.equals(classNames[0])) {
+            Method get = adInfo.getClass().getMethod("get", long.class, TimeUnit.class);
+            adInfo = get.invoke(adInfo, 5, TimeUnit.SECONDS);
```
e7mac, line 218: why do we need to do this timeout? Is this us handling the exception instead of passing that over?

milos1290: The return type is future, need to wait for few seconds to get the result back

PR #395, opened 2020-05-12
LP-11310 - add google play url support in web interstitials

hborisoff:
What              | Where/Who
JIRA Issue        | [LP-11310](https://leanplum.atlassian.net/browse/LP-11310)

Handle URLs to Google Play apps and open with intent in Web Interstitial message.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/messagetemplates/BaseMessageDialog.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/messagetemplates/BaseMessageDialog.java
@@ -24,13 +24,16 @@ package com.leanplum.messagetemplates;
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.Dialog;
+import android.content.ActivityNotFoundException;
 import android.content.Context;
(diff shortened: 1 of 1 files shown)


AndroidSDKCore/src/main/java/com/leanplum/messagetemplates/BaseMessageDialog.java
```diff
@@ -456,6 +459,23 @@ public boolean shouldOverrideUrlLoading(WebView wView, String url) {
...
+          try {
+            Intent intent = new Intent(Intent.ACTION_VIEW);
+            intent.setData(uri);
+            wView.getContext().startActivity(intent);
+            return true;
+          } catch (ActivityNotFoundException e) {
+            // Missing Google Play. Opening link directly
+            String host = uri.getHost();
+            String query = uri.getQuery();
+            wView.loadUrl("https://play.google.com/store/apps/" + host + "?" + query);
```
milos1290, line 475: I don't think we should open link in our IAM if we fail to redirect.

PR #400, opened 2020-06-22
Add snapshot testing framework

hborisoff:
What              | Where/Who
JIRA Issue        | [E2-2127](https://leanplum.atlassian.net/browse/E2-2127)

## Background
This PR is adding support for snapshot tests in SDK.
It uses the Facebook developed library - [screenshot-tests-for-android](https://facebook.github.io/screenshot-tests-for-android/).

Android emulator with API 22 (ARM architecture) is chosen after testing other emulator versions. Added code to create and start it in the Travis build script.
Wrote AndroidSDKTests/README.md on how to execute and record tests.

Tests for most of the message templates are written but there are couple of them which are not testable with current implementation - Alert and Confirm.


AndroidSDKTests/build.gradle
```diff
@@ -50,7 +62,7 @@ dependencies {
     implementation project(':AndroidSDKPush')
     implementation project(':AndroidSDKLocation')
     implementation project(':AndroidSDKFcm')
-    implementation "androidx.appcompat:appcompat:${APPCOMPAT_LIBRARY_VERSION}"
+    implementation "androidx.appcompat:appcompat:1.2.0-beta01" // WebView in Emulator with API22 is failing with exception because of appcompat version. In 1.2.0-beta01 it is not reproduced.
```
e7mac, line 65: should we upgrade the `APPCOMPAT_LIBRARY_VERSION` directly? I've definitely had issues with mixing versions before

hborisoff: Do you remember what kind of issues - runtime or compile?

Because I haven't seen any by far. Of course it is better to keep one version for all modules but I would prefer not to use beta versions in our published packages. This will cause the clients to download that dependency too and they might not be ok.

There is a task ([LP-11376](https://leanplum.atlassian.net/browse/LP-11376)) in backlog from yesterday that will address that version difference and will do it as soon as there is a release version available.

PR #402, opened 2020-06-29
Fix start callbacks to be invoked on failure.

hborisoff:
What              | Where/Who
JIRA Issue        | [LP-11342](https://leanplum.atlassian.net/browse/LP-11342)

## Background
In some cases the start callbacks were not invoked when exception is thrown during HTTP call.
After this fix all waiting callbacks will be invoked on a failure.

Two unit tests are added to verify that the callbacks are invoked on failure - one is for common exception from HTTP framework (this includes the SSL handshake exception), second is when response code is 500 (internal server error).


AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
```diff
@@ -361,7 +361,7 @@ private static boolean attachApiKeys(Map<String, Object> dict) {
    * @param responseBody JSONObject with response body from server.
    * @param error Exception.
    */
-  protected void parseResponseBody(JSONObject responseBody, Exception error) {
+  protected void triggerCallbackManager(JSONObject responseBody, Exception error) {
```
e7mac, line 364: are we actually triggering a callback manager here?

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -725,9 +731,10 @@ private static void handleStartResponse(final JSONObject response) {
         // Load the variables that were stored on the device from the last session.
         VarCache.loadDiffs();
 
-        triggerStartResponse(false);
       } catch (Throwable t) {
         Util.handleException(t);
+      } finally {
+        triggerStartResponse(success);
```
e7mac, line 737: are we triggering a start response or start request? Maybe its an opportunity to improve our naming?

hborisoff: We have the following methods in Leanplum.java:

addStartResponseHandler
removeStartResponseHandler
triggerStartResponse

There is a pattern for the names and If I change one method (triggerStartResponse) then I should change all of them and brake the SDK backward compatibility because the add and remove methods are public. I don't think we should do that at the moment.

AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
hborisoff: I changed the name to that one because the name does not show one of the main things that this method is doing - invoking the callbacks from the requests in the response, or error callbacks for all of the callbacks in the manager.

If you have a better suggestion I could change it.

PR #403, opened 2020-07-07
Remove 'background' parameter.

hborisoff:
What              | Where/Who
JIRA Issue        | [LP-11374](https://leanplum.atlassian.net/browse/LP-11374)

## Background
The 'background' parameter is not used based on the API logs and also the SDK cannot be used in the background.


AndroidSDKCore/src/main/java/com/leanplum/internal/RequestOld.java
```diff
@@ -454,7 +454,7 @@ protected RequestsWithEncoding getRequestsWithEncodedStringStoredRequests(double
         requestsToSend = new ArrayList<>(0);
       } else {
         unsentRequests = getUnsentRequests(fraction);
-        requestsToSend = removeIrrelevantBackgroundStartRequests(unsentRequests);
+        requestsToSend = new ArrayList<>(unsentRequests);
```
nzagorchev, line 457: Does this ensure that start action will not be called more than once if it has failed? This might probably happen at the next start of the app only.

hborisoff: The start action is sent by sendIfConnected so every time the app is opened all batched events are sent. If I stop the data on the device and try to simulate multiple start events by killing the app and running it again it produces the following sequence of events:

start, pauseSession, start, pauseSession, start, pauseSession and so on.

Cannot reproduce batch with multiple start events one after another so decided to fully remove this removeIrrelevantBackgroundStartRequests method.

hborisoff:
After further investigation the reason for 'background' flag in Android SDK is that the start call is initiated from the Application class of the app. This Application class implements the singleton pattern and is created whenever the OS does need to create Activity, Service, BroadcastReceiver, Provider or other.
For example when device receives push notification it creates the service that is responsible for handling that push notification without showing any UI. This is one of the cases when the SDK starts with background=true state.

Will close this PR as it is braking the implementation.

PR #404, opened 2020-07-10
Network layer refactoring

hborisoff:
What              | Where/Who
JIRA Issue        | [SDK-5](https://leanplum.atlassian.net/browse/SDK-5)

## Background
Tried to not change a lot of the logic in the code and to be as close as possible to the documentation and to iOS.

Made some things differently:

- RequestBuilder - implementation of builder pattern for the requests. Usage (placeholders are uppercased):
`RequestBuilder.with"NAME_OF_ACTION"Action().andParam(PARAM_NAME, PARAM_VALUE).andParams(MAP_OF_PARAMS).create()`

Kept the RequestFactory.create() for the unit tests. A lot of them depend on code that is overriding the default factory. This might be removed in future.

- RequestSender.sendRequests() does not depend on Request object parameter.
- Added FileTransferManager where most of the code downloading and uploading files is placed
- Requesting interface is removed
- RequestOld becomes Request and the static fields (appId, accessKey, etc) are moved to APIConfig


AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018, Leanplum, Inc. All rights reserved.
+ * Copyright 2013, Leanplum, Inc. All rights reserved.
```
milos1290, line 2: Update copyright year

AndroidSDKCore/src/main/java/com/leanplum/internal/Request.java
```diff
@@ -23,51 +23,81 @@
...
-  public Request(String httpMethod, String apiMethod, Map<String, Object> params) {
+  public Request(String httpMethod, String apiAction, Map<String, Object> params) {
     this.httpMethod = httpMethod;
-    this.apiMethod = apiMethod;
-    this.params = params;
-    this.countAggregator = Leanplum.countAggregator();
+    this.apiAction = apiAction;
+    this.params = params != null ? params : new HashMap<>();
+    // Check if it is error and here was SQLite exception.
+    if (RequestBuilder.ACTION_LOG.equals(apiAction) && LeanplumEventDataManager.sharedInstance().willSendErrorLogs()) {
```
milos1290, line 57: This should be moved to request sender instead of checking it here. 
We can do it another PR

hborisoff: It is better to do this in another PR.

PR #410, opened 2020-08-10
Refactor IAM templates (part 2)

hborisoff:
What              | Where/Who
JIRA Issue        | [SDK-71](https://leanplum.atlassian.net/browse/SDK-71)

## Background
- Moved most of the 'defineAction' code to MessageTemplates.java and did leave only 'createActionArgs' and 'showMessage' methods per message instead of the 'register' method.
- Introduced MessageTemplateConstants.java to reduce the size of MessageTemplates.java.
- Renamed the message classes to be distinguishable enough from the other classes in the package (HTMLTemplate becomes RichMessage, OpenURL becomes OpenUrlAction, appended 'Message' suffix to the other messages)


hborisoff:
AndroidSDK/out.map - this file needs to be removed from the repository and added to .gitignore. Will do it in another PR.

PR #417, opened 2020-08-27
Add snapshot test for banner message

hborisoff:
What              | Where/Who

## Background
Added missing snapshot test about banner messages.


hborisoff:
Will create new PR after resolving the issue to not spam with emails.

PR #425, opened 2020-10-19
Refactor batching of requests

hborisoff:
What              | Where/Who
JIRA Issue        | [SDK-140](https://leanplum.atlassian.net/browse/SDK-140)

## Background
There are several changes to the batching functionality:

1. Removed `LeanplumDataBaseManager.migrateFromSharedPreferences` as the migration was done several years ago and there aren't any clients using such old sdk version.

2. Removed `synchronized (Request.class)` synchronisation code from several places. All of the code was running synced except the `LeanplumDataBaseManager.migrateFromSharedPreferences` method, which is now removed.

3. Moved the old inner class `RequestsWithEncoding` to a separate `RequestBatch.java`.

4. Moved the batching codebase from `RequestSender` to a `RequestBatchFactory` class. It contains several new public methods:
- `createErrorBatch` - called if there is DB error
- `createNextBatch` - gets the next batch of events from DB
- `deleteFinishedBatch` - deletes the events from DB

5. UUID parameter
After discussion with backend guys the request UUID param needs to be persisted with the requests in the DB because on server-side there is a 'dedup' logic that compares them by **uuid** and **reqId** parameters, and ignores the duplicates.

Moved all UUID related code to `RequestUuidHelper.java`. It contains several public methods:
- `deleteUuid` - deletes the current uuid from shared prefs
- `saveNewUuid` - saves a new uuid in shared prefs
- `loadUuid` - loads the current uuid from shared prefs
- `attachUuid` - attaches uuid parameter to the event
- `attachNewUuid(List)` - attaches new uuid parameter to every event in the list. The new uuid is not saved in shared prefs


AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
```diff
@@ -205,13 +205,18 @@ public void onCreate(SQLiteDatabase db) {
...
-      } catch (Throwable t) {
-        Log.e("Cannot move old data from shared preferences to SQLite table.", t);
-        Log.exception(t);
-      }
+      OperationQueue.sharedInstance().addOperation(new Runnable() {
+        @Override
+        public void run() {
+          // Migrate old data from shared preferences.
+          try {
+            migrateFromSharedPreferences(db);
```
milos1290, line 213: I think we can remove migrations. It was added when we switched from shared prefs to sqlite, but i dont think anyone is using SDK that old anymore.

milos1290 (changes requested):

AndroidSDKCore/src/main/java/com/leanplum/internal/LeanplumEventDataManager.java
hborisoff: Done

milos1290 (approved):

PR #428, opened 2020-11-05
Add customizable interval in RequestSenderTimer

hborisoff:
What              | Where/Who
JIRA Issue        | [SDK-163](https://leanplum.atlassian.net/browse/SDK-163)

## Background
Added 3 options for customisation of events upload interval - 5, 10, or 15 minutes.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/EventsUploadInterval.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/EventsUploadInterval.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2020, Leanplum, Inc. All rights reserved.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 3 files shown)


nzagorchev (approved):

AndroidSDKCore/src/main/java/com/leanplum/EventsUploadInterval.java
```diff
@@ -0,0 +1,53 @@
...
+ * under the License.
+ */
+
+package com.leanplum;
+
+/**
+ * The enumeration represents time interval to periodically upload events to server.
+ * Possible values are 5, 10, or 15 minutes.
+ */
+public enum EventsUploadInterval {
```
dejan2k, line 28: What do you think about:
`PeriodicSyncInterval`

hborisoff: It is only about uploading data to server. As I understand what 'sync' means is that the data on device will be synced with data on server, but that will happen only on Start request, if you don't have the Start event in the DB it won't download data from server.

dejan2k: Yeah actually we talked about that, I was thinking about `PeriodicUploadInterval` but now when I wrote it probably `EventsUploadInterval` sounds better

dejan2k (approved):

milos1290 (approved):

PR #434, opened 2020-12-03
Add method for downloading inbox messages

hborisoff:
What              | Where/Who
JIRA Issue        | [SDK-232](https://leanplum.atlassian.net/browse/SDK-232)

## Implementation
Adding method to allow clients to download inbox messages without the need to call `forceContentUpdate`.

Two methods are added to public interface:

`LeanplumInbox.downloadMessages()` - Triggers downloading of messages. `LeanplumInbox.syncedCallbacks` will be invoked when operation is done.

`LeanplumInbox.downloadMessages(InboxSyncedCallback)` - Triggers downloading of messages and adds the callback to `LeanplumInbox.syncedCallbacks`. When downloading is complete the callback is removed from the list.

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
@@ -25,12 +25,8 @@ import android.content.Context;
 import android.location.Location;
 import android.text.TextUtils;
 
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 import com.leanplum.ActionContext.ContextualValues;
 import com.leanplum.callbacks.ActionCallback;
-import com.leanplum.callbacks.InboxChangedCallback;
-import com.leanplum.callbacks.InboxSyncedCallback;
 import com.leanplum.callbacks.MessageDisplayedCallback;
 import com.leanplum.callbacks.RegisterDeviceCallback;
 import com.leanplum.callbacks.RegisterDeviceFinishedCallback;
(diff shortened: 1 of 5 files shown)


nzagorchev (approved):
Nice callback addition.

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -2080,6 +2084,24 @@ public void error(Exception e) {
...
+    downloadInboxMessages(null);
+  }
+
+  /**
+   * Forces downloading of inbox messages from the server. After messages are downloaded the
+   * appropriate callbacks will fire.
+   *
+   * @param callback The callback to invoke when messages are downloaded.
+   */
+  public static void downloadInboxMessages(@Nullable InboxSyncedCallback callback) {
```
milos1290, line 2101: Should we move those 2 methods to be accessible only from inbox() instance ?

milos1290 (changes requested):

AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
hborisoff: This is a good question, I thought that it would be better to add them in Leanplum class, but now I see that it will be better to move them to LeanplumInbox, because there already are several public methods for registering the callbacks.
I will update the PR.

milos1290 (approved):

PR #460, opened 2021-06-28
Add handler for runActionNamed

hborisoff:
What              | Where/Who
JIRA Issue        | [SDK-455](https://leanplum.atlassian.net/browse/SDK-445)

This change enables callback to be invoked when in-app message actions are run.
For example - `Accept action`, `Select button 1 action`.

Added `Dismiss action` to be run for some of the in-app templates, where appropriate.

Example usage:

```
Leanplum.onAction("Center Popup", new ActionCallback() {
    @Override
    public boolean onResponse(ActionContext context) {
        context.setActionNamedHandler(new ActionCallback() {
            @Override
            public boolean onResponse(ActionContext context) {
                String actionName = context.actionName(); // "Accept action"
                String parentActionName = context.getParentContext().actionName(); // "Center Popup"
                return true;
            }
        });
        return true;
    }
});
```

### DIFF
--- a/AndroidSDKCore/src/main/java/com/leanplum/ActionContext.java
+++ b/AndroidSDKCore/src/main/java/com/leanplum/ActionContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014, Leanplum, Inc. All rights reserved.
+ * Copyright 2021, Leanplum, Inc. All rights reserved.
  *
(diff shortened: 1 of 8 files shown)


nzagorchev (approved):
LGTM.
Tested dismiss actions and responder.

PR #464, opened 2021-07-21
Fix NullPointerException when message is missing

hborisoff:
What              | Where/Who


hborisoff:
Closing PR because issue was reproduced in older version and it is not related to current SDK.

PR #465, opened 2021-08-10
Integrate FCM 22.0.0.

JayShortway:
What              | Where/Who
JIRA Issue        | N/A
People Involved   | Just me

## Background
`firebase-messaging` 22.0.0 deprecates the [`FirebaseInstanceId`](https://firebase.google.com/docs/reference/android/com/google/firebase/iid/FirebaseInstanceId) class, which is used by `LeanplumFcmProvider`. This causes a crash in apps that implement `firebase-messaging` 22.0.0 together with Leanplum 5.7.0. 
  
## Implementation
This PR implements `firebase-messaging` 22.0.0 by changing the references to `FirebaseInstanceId` in `LeanplumFcmProvider` to `FirebaseMessaging`.

## Testing steps
All tests should still pass, including `LeanplumPushServiceTest`. Moreover, this change successfully sets a push token for our device, visible on Leanplum's web client. We also receive push notifications as expected.  

Unfortunately, this change is not backwards-compatible with versions of `firebase-messaging` before 22.0.0. Please let me know if you want to accommodate for that, and the preferred way of doing so. I will try to implement your suggestions.

### DIFF
--- a/AndroidSDKFcm/build.gradle
+++ b/AndroidSDKFcm/build.gradle
@@ -34,7 +34,7 @@ dependencies {
     api project(':AndroidSDKPush')
 
     // Provided dependencies are optional dependencies and will not show up in pom file.
-    compileOnly('com.google.firebase:firebase-messaging:[17.3.4, 21.1.0]') {
+    compileOnly('com.google.firebase:firebase-messaging:22.0.0') {
         exclude module: 'support-v4'
     }
 }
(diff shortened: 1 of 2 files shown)


hborisoff:
@JayShortway, thank you for your PR. We are aware of the change that needs to be done, but the latest FCM package, supporting these new methods, is still not old enough to require our clients to update to it. Check my PR #454. Currently we support FCM up to 21.1.0.
We'll update it to use the new methods in several months.

PR #478, opened 2021-12-09
Add `setLocale` method

batadamnjanovic:
Introduce `setLocale(locale: String)` in order to set a custom locale string.

Problem with this approach is that `setLocale(...)` needs to be called before `Leanplum.start(...)` method.


hborisoff:
Hi @batadamnjanovic, what is your idea in adding setter method, because you can easily change the device's language if you want to test locales?

batadamnjanovic:
Hi @hborisoff, this is similar to Web: https://github.com/Leanplum/Leanplum-JavaScript-SDK/pull/140 and iOS: https://github.com/Leanplum/Leanplum-iOS-SDK/pull/467.

We have predefined language codes that we need to send with every event.

hborisoff (approved):

hborisoff:
@batadamnjanovic Yesterday I was working on a [release 5.8.2](https://github.com/Leanplum/Leanplum-Android-SDK/releases/tag/5.8.2) containing a bug fix and included your change too.

PR #480, opened 2022-01-13
Add RichStartCallback

opengamer29:


hborisoff:
@opengamer29 Can you check my [change](https://github.com/Leanplum/Leanplum-Android-SDK/commit/4bdb7d7566e5d4bf19f2f02f586a8a62338934cf) about the exception listener to verify that it would help you, because I'm planing a release these days and I want to include it?

PR #561, opened 2024-04-02
Fix userId occasionally reset on app launch

vasct:
JIRA Issue        | [SDK-3718](https://wizrocket.atlassian.net/browse/SDK-3718)


AndroidSDKCore/src/main/java/com/leanplum/Leanplum.java
```diff
@@ -900,8 +900,10 @@ private static void handleStartResponse(final JSONObject response) {
           }
         }
 
-        String token = response.optString(Constants.Keys.TOKEN, null);
-        APIConfig.getInstance().setToken(token);
+        String token = response.optString(Constants.Keys.TOKEN);
```
vasct, line 903: optString returns empty string as default. Passing ```null``` as default generates a lint warning because the parameter is marked as ```@NonNull```

nzagorchev (approved):

